PR #3, opened 2019-06-17
README: avoid inline `uuidgen`

cohuck:
Calling `uuidgen` inline with create-mdev makes it hard to figure
out which device was created if mdevctl is invoked concurrently
multiple times. Better make the invocation of uuidgen explicit in
the example.

Signed-off-by: Cornelia Huck <cohuck@redhat.com>


awilliam:
Thanks Connie!

PR #15, opened 2019-12-18
validate attribute names

cohuck:
Try to defend against breaking out of the sysfs hierarchy via
the attribute names, e.g. '../../../../../../../../boo'.

Signed-off-by: Cornelia Huck <cohuck@redhat.com>

### DIFF
--- a/mdevctl
+++ b/mdevctl
@@ -23,6 +23,20 @@ jsonify() {
     echo "\"$1\""
 }
 
+validate_attr() {
+    # first arg: expected root, second arg: attr to check
+    if [ ! -w "$1/$2" ]; then
+       echo "Attribute $2 cannot be set" >&2
+       return 1
+    fi
+    if [ $(realpath --relative-base "$1" "$1/$2") == "$2" ]; then
+        return 0
+    else
+        echo "Invalid attribute $2" >&2
+        return 1
+    fi
+}
+
 del_config_key() {
     key="$1"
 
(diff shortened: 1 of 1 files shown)


awilliam:
This seems to work, it prevents the obvious attempts to get into parent directories with ../ and also prevents walking symlinks from the device directory, which I think is also valid.  A nit though is that 'modify --addattr' will happily store bogus attributes in the mdev definition, which only fail when we try to apply them.  I don't think this is fully solvable though as we don't require the mdev to be present as we're defining or modifying its definition.  Would it be worthwhile though to at least verify an added attribute isn't gratuitously escaping?  For instance, if we used "." rather than the sysfs device path for a similar realpath test when adding attributes, we could catch the obvious ../ escape, but not more intricate sysfs symlink loops that might still allow escaping sysfs entirely.  Using "." also has problems as it depends on the user's location when running mdevctl (ex. / nulifies ../).  Is it worthwhile to try to catch the more obvious escapes when adding attributes even if they gets caught by the test added here when we try to apply it?

cohuck:
> A nit though is that 'modify --addattr' will happily store bogus attributes in the mdev definition, which only fail when we try to apply them. I don't think this is fully solvable though as we don't require the mdev to be present as we're defining or modifying its definition.

Yes, I also looked at that code path and came to the same conclusion; we cannot even verify that the attribute exists in the first place. Not sure if it is that important; we need to catch problems while applying anyway, as the user can simply drop in a config file.

> Would it be worthwhile though to at least verify an added attribute isn't gratuitously escaping? For instance, if we used "." rather than the sysfs device path for a similar realpath test when adding attributes, we could catch the obvious ../ escape, but not more intricate sysfs symlink loops that might still allow escaping sysfs entirely. Using "." also has problems as it depends on the user's location when running mdevctl (ex. / nulifies ../). Is it worthwhile to try to catch the more obvious escapes when adding attributes even if they gets caught by the test added here when we try to apply it?

I'm not sure if checking against '.' is a good idea for the reasons you gave.

One thing we might want to do instead is prohibiting using '../' in attributes in general...

PR #22, opened 2020-05-19
Fix the Makefile so that it works on CentOS 7 as well

eskultety:
Currently the Makefile uses some Git >2.0 features not available on CentOS-7, making it unable to build the rpms from source. By using the common Linux OS binaries, this can be fix in a straightforward manner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ archive: files mdevctl.spec
 mdevctl.spec: tag mdevctl.spec.in files
 	sed -e 's:#VERSION#:$(VERSION):g' < mdevctl.spec.in > mdevctl.spec
 	PREV=""; \
-	for TAG in `git tag --sort=version:refname | tac`; do \
+	for TAG in `git tag | sort -Vr`; do \
 	    if [ -n "$$PREV" ]; then \
 	    git log --format="- %h (\"%s\")" $$TAG..$$PREV >> mdevctl.spec; \
 	    fi; \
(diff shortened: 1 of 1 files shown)


awilliam:
Thanks for the patches Erik.  Do you see it as a requirement to support mdevctl on Centos7?  There's at least a uevent that's not in the RHEL7 kernel for triggering mdevctl udev rules, I don't think we plan to backport that, or provide mdevctl on RHEL7 given its development phase.  Therefore I wonder if it really makes sense to use more convoluted methods to support features built into the current tool stack.

eskultety:
> Thanks for the patches Erik. Do you see it as a requirement to support mdevctl on Centos7? There's at least a uevent that's not in the RHEL7 kernel for triggering mdevctl udev rules, I don't think we plan to backport that, or provide mdevctl on RHEL7 given its development phase. Therefore I wonder if it really makes sense to use more convoluted methods to support features built into the current tool stack.
These patches were merely a result of testing some libvirt mdev patches on CentOS-7 and I prefer not having to do make install. I didn't know that we're missing a uvent, so like you say, it doesn't make much sense to pursue these changes given the situation. However, maybe the first patch reverse sorting tags could still be an enhancement in terms of readability by simply using "sort -Vr", I'll leave it up to you Alex :).

cohuck:
I'll go ahead and close this. With the Rust rewrite, we currently support building on RHEL 8  and Fedora 34+. I think CentOS 7 is out of scope for this -- although I'd certainly not be opposed to making other distributions work, if it does not turn out too ugly :)

Issue #24, opened 2020-06-26
Conflict with short options for 'auto' and 'available'


etrunko:
Both use '-a', and because 'available' is tested later, the code will never reach it. Reported by coverity scan.

```
Defect type: SHELLCHECK_WARNING 
1. /usr/sbin/mdevctl:543:9: warning: This pattern never matches because of a previous pattern. [SC2222]
#   541|               shift 1
#   542|               ;;
#   543|->         -a|--available)
#   544|               available=y
#   545|               shift 1
```
```
Defect type: SHELLCHECK_WARNING 
1. /usr/sbin/mdevctl:531:9: warning: This pattern always overrides a later one. [SC2221]
#   529|               shift
#   530|               ;;
#   531|->         -a|--auto)
#   532|               auto=y
#   533|               shift 1
```

PR #25, opened 2020-06-26
Remove never used 'available' option

etrunko:
As reported by coverity scan:

Defect type: SHELLCHECK_WARNING
1. /usr/sbin/mdevctl:544:13: warning: available appears unused. Verify use (or export if used externally). [SC2034]

This also fixes a conflict between `auto` and `available` short options,
with both using -a for it.

closes: #24

Signed-off-by: Eduardo Lima (Etrunko) <etrunko@redhat.com>

### DIFF
--- a/mdevctl
+++ b/mdevctl
@@ -540,10 +540,6 @@ while true; do
             defined=y
             shift 1
             ;;
-        -a|--available)
-            available=y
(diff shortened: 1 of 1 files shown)


awilliam:
Thanks!

Issue #26, opened 2020-07-10
Use standart path to bash


shaba:
Fix /usr/bin/bash to /bin/bash in mdevctl shebang.

cohuck:
Is /bin/bash really standard? On newer Linux distros, /bin is a symlink to /usr/bin, and I think /usr/bin/bash is/was the default location on some non-Linux systems.

If /usr/bin/bash is really a problem, using /usr/bin/env would seem like a better solution.

shaba:
mdevctl can be used in distributives with /usr as mount point. Then /usr/bin/env is bad idea.
Distributives with merged /bin and /usr/bin have a symlink /bin -> /usr/bin.
I think use in shebang old style /bin/sh or /bin/bash ise best choice.

cpaelzer:
I thought it is only me/us that are different and [silently changed the path for Debian/Ubuntu](https://sources.debian.org/src/mdevctl/0.61-1/debian/patches/debianize-paths.patch/).
But I agree to @shaba that '/bin/bash' should be the most reliable and common one.

Issue #14, opened 2019-12-11
mdevctl should have a way to print out its version


cohuck:
We probably should add something like mdevctl --version to print out the release number and date.

awilliam:
Resolved in 0.78 by b45d33dfd36cb5baffcb742b8dbf645882847b5d, 'mdevctl version'

PR #29, opened 2020-09-17
Fixes for #14 #26 and a few minor things

cpaelzer:
Hi,
while I was merging 0.69 for Debian/Ubuntu today I was not only confused by the version number jumps (see IRC question) but also saw that there are a few - what seems to be trivial - issues we could fix. Of course makefile magic always ends up more complex than it should be, but still I thought I'll provide a PR with those fixes for your consideration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ SBINDIR=$(PREFIX)/sbin
 CONFDIR=/etc/mdevctl.d
 MANDIR=$(PREFIX)/share/man
 NAME=mdevctl
-VERSION=0.$(shell git rev-list --count HEAD)
+LASTVERSION:=$(shell git tag --sort=version:refname --merged| tail -n 1 | sed 's/\([0-9]*\.\)\([0-9]*\)/\2/g')
+VERSION:=0.$(shell echo $$(($(LASTVERSION)+1)))
 NVFMT=$(NAME)-$(VERSION)
 
 files: mdevctl 60-mdevctl.rules mdevctl.8 \
--- a/Makefile
+++ b/Makefile
@@ -48,4 +48,14 @@ clean:
 	rm -f mdevctl.spec *.src.rpm noarch/*.rpm *.tar.gz
 
 tag:
-	git tag -l $(VERSION) | grep -q $(VERSION) || git tag $(VERSION)
+	@if git describe --exact-match 2>/dev/null 1>&2; then \
+		echo "Current commit is already tagged as:"; \
+		git describe --exact-match; \
+		false; \
+	else \
+		if ! git tag -l $(VERSION) | grep $(VERSION); then \
+			sed -i -e 's/\(^\s*echo "\$$0 version \)[0-9]*\.[0-9]*/\1$(VERSION)/g' mdevctl; \
+			git commit -s -m "update version" mdevctl; \
+			git tag $(VERSION) -a -m "$(VERSION)"; \
+		fi; \
+	fi
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ archive: files mdevctl.spec
 mdevctl.spec: tag mdevctl.spec.in files
 	sed -e 's:#VERSION#:$(VERSION):g' < mdevctl.spec.in > mdevctl.spec
 	PREV=""; \
-	for TAG in `git tag --sort=version:refname | tac`; do \
+	for TAG in `git tag --sort=version:refname --merged | tac`; do \
 	    if [ -n "$$PREV" ]; then \
 	    git log --format="- %h (\"%s\")" $$TAG..$$PREV >> mdevctl.spec; \
 	    fi; \
--- a/mdevctl
+++ b/mdevctl
@@ -312,7 +312,7 @@ unique_uuid() {
 
 usage() {
     cat >&2 <<EOF
-Usage: $(basename $0) {COMMAND} [options...]
+Usage: $(basename $0) [--help || --version] {COMMAND} [options...]
 
 Available commands:
 define		Define a config for an mdev device.  Options:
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -4,7 +4,8 @@ SBINDIR=$(PREFIX)/sbin
 CONFDIR=/etc/mdevctl.d
 MANDIR=$(PREFIX)/share/man
 NAME=mdevctl
-VERSION=0.$(shell git rev-list --count HEAD)
+LASTVERSION:=$(shell git tag --sort=version:refname --merged| tail -n 1 | sed 's/\([0-9]*\.\)\([0-9]*\)/\2/g')
```
cohuck, line 7: This way indeed should produce saner version numbers.

While we're at it: Does the 0.<n> scheme make much sense if we always just increment anyway? Do we have a threshold for bumping the first part to 1, or would it make sense to switch to a single incremental number instead? I'm not sure if we want to do something like semantic versioning at all. @awilliam

mdevctl
```diff
@@ -431,6 +434,10 @@ case ${1} in
     #
     # User commands
     #
+    --version|-v)
+        echo "$0 version 0.69"
```
cohuck, line 438: Eternal version 0.69?

cpaelzer: No @cohuck as soon as you `make tag` or `make archive` (which implicitly tags) this will get updated.

Makefile
cpaelzer: @cohuck I thought it is just 0.x being new and didn't think about changing the scheme, but surely while touching how it increments we can think if we want to change.

IMHO The transition from 0.x to 1.x usually indicates some implication about maturity or "planned features reached".
Not sure if you had a specific set of features or use cases in mind to wait for 1.x @awilliam.

mdevctl
cpaelzer: We could throw in a comment here to indicate that changes will affect the sed rules in the Makefile maybe?

cohuck: As you can see from my comment, adding an explanation would be useful :)

cpaelzer: ok - added to the proposed branch @cohuck

Makefile
awilliam: I don't see what's so mysterious or confusing about the current versioning scheme (sorry, I don't know what IRC question this is referencing).  Version = commit index.  Is this change simply to be able to implement a -version arg?  Version numbers are largely meaningless other than they need to be monotonically increasing.  The current scheme provides that and implicitly indicates the number of commit between versions.  What does it matter if/when we reach 1.0 or beyond?

awilliam: Regarding 0.x vs 1.x, I see it as a mechanism to potentially introduce forks that might break compatibility.  My intention would be to continue on with 0.x numbering indefinitely, assuming we don't find a reason to break compatibility.

cohuck (commented):
@awilliam Does this change still make sense, even if we go with your approach to versioning?

cohuck (commented):
@awilliam Do we want to merge this? (Not sure how to pick individual commits on github.)

Makefile
```diff
@@ -48,4 +48,14 @@ clean:
 	rm -f mdevctl.spec *.src.rpm noarch/*.rpm *.tar.gz
 
 tag:
-	git tag -l $(VERSION) | grep -q $(VERSION) || git tag $(VERSION)
+	@if git describe --exact-match 2>/dev/null 1>&2; then \
```
cohuck, line 51: Looking at this again... @awilliam , should we add some kind of sanity check if we go with your version?

awilliam: probably... what do you suggest?

awilliam:
> @awilliam Do we want to merge this? (Not sure how to pick individual commits on github.)

Sorry, I'm not sufficiently adept at github reviews to know specifically what either of these comments are referring to...

Makefile
cohuck: Basically, something similar to this change here: check via git describe whether we already tagged a version here, and bail out if yes.

cohuck:
> > @awilliam Do we want to merge this? (Not sure how to pick individual commits on github.)
> 
> Sorry, I'm not sufficiently adept at github reviews to know specifically what either of these comments are referring to...

Ok, "this" was referring to the commit "avoid other tags spilling into the .spec changelog". I'm not adept at github reviews, either :(

Makefile
awilliam: That's essentially what [ $(MDEVCTL_VER) == $(REVLIST_VER) ] does, it tests that the version reported by mdevctl matches the number of commits to HEAD and only tags if they're different.  Therefore repeatedly running 'make tag' should not perpetually create new tags.

cohuck: Fair enough.

awilliam:
> > > @awilliam Do we want to merge this? (Not sure how to pick individual commits on github.)
> > 
> > 
> > Sorry, I'm not sufficiently adept at github reviews to know specifically what either of these comments are referring to...
> 
> Ok, "this" was referring to the commit "avoid other tags spilling into the .spec changelog". I'm not adept at github reviews, either :(

Sure, the --merged addition seems fine.

Makefile
cpaelzer: Yeah would work with me as well - thanks!

Makefile
cpaelzer: The discussion went on and we concluded I think, but since this conversation is still open I wanted to thank for the explanation and close it (for a better overview).

cpaelzer:
To be sure, do you cherry-pick and add commits yourself for what we concluded @cohuck / @awilliam ?
Or do you want me to rework/rebase something here (and if so what should I prep for you)?

cohuck:
> To be sure, do you cherry-pick and add commits yourself for what we concluded @cohuck / @awilliam ?
> Or do you want me to rework/rebase something here (and if so what should I prep for you)?

Given how I'm struggling with the interface here, a reworked pull request with just the bash path and the --merged changes seems like the easiest way to do this. This should be independent of #31 .

awilliam:
Cherry picked the bash path and --merged commits manually from pr.  Thanks!

cpaelzer:
Thank you @awilliam

PR #31, opened 2020-11-13
Bug fix + version support

awilliam:
1) Fix an obvious bug where the path we call when a parent device appears only creates the mdev without applying attributes.

2) Add version command support.  The 'tag' make target will now modify the version variable in the mdevctl script, commit the change, and add the matching tag.  This is not too unlike kernel versions, where the Makefile is updated and a tag added to that commit with a matching version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,10 @@ SBINDIR=$(PREFIX)/sbin
 CONFDIR=/etc/mdevctl.d
 MANDIR=$(PREFIX)/share/man
 NAME=mdevctl
-VERSION=0.$(shell git rev-list --count HEAD)
-NVFMT=$(NAME)-$(VERSION)
+MDEVCTL_VER=$(shell ./mdevctl version)
+REVLIST_VER=0.$(shell git rev-list --count HEAD)
+NEXT_VER=0.$(shell echo $$(( $(shell git rev-list --count HEAD) + 1 )) )
+NVFMT=$(NAME)-$(REVLIST_VER)
 
 files: mdevctl 60-mdevctl.rules mdevctl.8 \
 	Makefile COPYING README.md mdevctl.spec.in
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,7 +17,7 @@ archive: files mdevctl.spec
 	gzip -f -9 $(NVFMT).tar
 
 mdevctl.spec: tag mdevctl.spec.in files
-	sed -e 's:#VERSION#:$(VERSION):g' < mdevctl.spec.in > mdevctl.spec
+	sed -e 's:#VERSION#:$(shell ./mdevctl version):g' < mdevctl.spec.in > mdevctl.spec
```
cohuck, line 20: Why not MDEVCTL_VER?

Makefile
```diff
@@ -47,4 +49,4 @@ clean:
 	rm -f mdevctl.spec *.src.rpm noarch/*.rpm *.tar.gz
 
 tag:
-	git tag -l $(VERSION) | grep -q $(VERSION) || git tag $(VERSION)
+	[ $(MDEVCTL_VER) == $(REVLIST_VER) ] || (sed -i "s/^version=.*/version=\"$(NEXT_VER)\"/" mdevctl && git add mdevctl && git commit -m "Automatic version commit for tag $(NEXT_VER)" && git tag $(NEXT_VER))
```
cohuck, line 52: Do we want -s with git commit?

(On a related note, do we want to require DCO?)

cohuck (approved):
LGTM.

Makefile
awilliam: Because after the tag target  version with either be MDEVCTL_VER or NEXT_VER depending on whether a new tag was created.

Makefile
awilliam: We don't currently require signed commits or tags, so this seems a bit out of scope for this change.  I always try to use DCO, though I wasn't sure it was really appropriate to insert a sign-off into an automated commit.  I'm not opposed to signing the tag, though I wonder if it's going to get in the way of unofficial builds or needs validation beyond what we have here.

cohuck: I think actually *signing* the tag is a different issue, no?

We may want to think about DCO and friends outside of this pull request; it just came to my mind.

Issue #27, opened 2020-07-13
Add notification events for add/remove device definitions


jonner:
A management app like libvirt would like to be able to monitor new or removed mediated device definitions. Currently, we have resort to something like periodically polling `mdevctl list --defined` to detect new or removed devices. It would be nice if we could instead listen for events indicating that a device is defined or undefined.

I envision something similar in concept to udev 'add' or 'remove' events for regular devices. Perhaps DBus could be the mechanism for announcing these events? I'm open to other ideas.

cohuck:
I'm wondering whether we can combine this with the 'callout' interface suggested in #20 -- we need a mechanism to invoke scripts etc. for that as well.

A post-create/post-remove callout should basically cover the add/remove events. If we do not introduce a more formal message via DBus, would the ability to invoke a custom script be already enough to process state changes?

jonner:
That's not a bad idea. It would probably work for what I need, depending on
exactly how it is implemented.

Issue #20, opened 2020-05-15
Support callouts at strategic points


cohuck:
Admins should be able to invoke external programs (or skripts) at various points during configuration/starting of devices. Examples would be:

- pre-config hook: does this new device look sane?
- post-config hook: does the state of the whole system look sane (i.e. definitions do not clash)?
- pre-start hook: are all requirements fulfilled that must be met before starting the device?

This would be useful to validate more complicated setups (e.g. setups with complex vfio-ap assignments.)

PR #32, opened 2021-04-08
Invoking External Scripts for Device Events

collinwalling:
This patch series introduces a method for mdevctl to invoke external scripts for device events and to broadcast notifications. See updates to README for more information.

Addresses #20 and #27.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
 ```
 
 See `mdevctl --help` or the manpage for more information.
+
+# Invoking External Scripts for Device Events
+
+Certain mediated devices may require additional operations or functionality,
+such as configuration checking or event reporting, before or after mdevctl
+executes a command. In order to remain device-type agnostic, mdevctl will
+"call-out" to external scripts to handle the extraneous work. These scripts
+are associated with a specific device type to perform any operations or
+additional functionality not handled within mdevctl. Additionally, external
+programs may wish to receive notifications of any action performed by mdevctl
+to e.g. respond to any device changes or keep device management in parallel.
+
+A call-out script is invoked at various points during an mdevctl command
+process and are categorized by an "event" paired with an "action", along
+with information regarding the mediated device. Two main event types are
+"pre" and "post", which are invoked before and after primary command
+execution respectively. The same call-out script is invoked for both events.
+
+A "notify" event script is invoked to report the status of mdevctl's
+command results. This may be used to signal external programs of changes
+made to a mediated device, or simply to assist with debugging efforts.
+
+A "get" event script is invoked for the define and list commands to acquire
+device attributes from sysfs.
+
+Essentially, the procedure in mdevctl looks like this:
+
+1. command-line parsing & setup
+2. invoke pre-command call-out
+3. primary command execution (e.g. start mdev / write device config)\*
+4. invoke post-command call-out\*
+5. invoke notifier\*
+
+\* step is skipped if 2 fails.
+
+## Script Parameters
+
+Each call-out and notifier script is invoked by mdevctl with the following
+parameters:
+
+**-e EVENT**
+: denotes the specific call-out or notification event
+- "pre" for pre-command call-out
+- "post" for post-command call-out
+- "notify" for notification scripts
+- "get" for acquiring device data
+
+**-a ACTION**
+: denotes what specific action the script is asked to do
+- for pre/post/notify events, this will be synonymous with an mdevctl command
+ (e.g. define, start)
+- for a get events, this will be "attributes"
+
+**-s STATE**
+: a trinary value defining the current state of mdevctl's command
+execution
+- none: mdevctl has yet to execute the command, or the pre-command
+call-out failed
+- success: mdevctl has completed the command successfully
+- failure: mdevctl has completed the command with an error
+
+**-u UUID & -p PARENT**
+: UUID and parent of the device.
+
+**stdin**
+: as standard input, the device's JSON configuration will be provided. This may
+represent:
+- a persistent device config created by the define command
+- a transient device config representing a sysfs device
+- an "in progress" config imposed by the modify command
+- a device config passed by the `--jsonfile` parameter.
+
+A script does not need to handle every action. As such, the script
+should take care to ignore any unrecognized / unsupported actions
+and allow mdevctl to carry on as normal.
+
+## Pre-command Event
+
+A pre-command event is invoked by mdevctl after any command-line parsing and
+setup, but before a command's execution (such as prior to writing the
+persistent device configuration, or prior to starting a device). For example,
+a script for a vfio_ap-passthrough device may check if the requested matrix is
+already in use before mdevctl can start the device.
+
+Errors reported by pre-command event scripts are disruptive. If a script
+reports an error, then mdevctl will exit early with an appropriate message
+and the command execution will not be performed (e.g. the device will not be
+defined or started). A notifier event will still be invoked in this case.
+
+This call-out will invoke a script with the "pre" `EVENT`, an `ACTION`
+reflecting the mdevctl command, a "none" `STATUS`, and a `UUID` and `PARENT`
+of the device.
+
+```
+echo $config | $script -e "pre" -a "start" -s "none" -u "f0bb71ac-9b72-4d2d-bbdb-67f41d3cd26e" -p "matrix"
+```
+
+## Post-command Event
+
+A post-command event is invoked by mdevctl after a command has executed (such
+as after a device configuration is written, or after a device has been started)
+to perform any additional steps that may be required for a device
+configuration. *The same script invoked for the pre-command call-out is also
+invoked for this call-out.*
+
+Errors reported by scripts during a post-command event are non-disruptive.
+
+This call-out will invoke a script with the "post" `EVENT`, an `ACTION`
+reflecting the mdevctl command, a `STATUS` reflecting the success/failure
+of mdevctl's primary command execution,and a `UUID` and `PARENT` of the
+device.
+
+```
+echo $config | $script -e "post" -a "start" -s "success" -u "f0bb71ac-9b72-4d2d-bbdb-67f41d3cd26e" -p "matrix"
+```
+
+## Notifier Events
+
+Notifier events are invoked by mdevctl to convey information to external
+listeners. For example, a script may signal to another program that an mdevctl
+command has executed successfully. A notifier always follows either a
+pre or post-command event.
+
+Errors reported by scripts during a notifier event are non-disruptive.
+
+A notifier call-out will invoke a script with the "notify" `EVENT`, an `ACTION`
+reflecting the mdevctl command, a `STATUS` reflecting the success/failure
+of mdevctl's primary command execution, or "none" if the pre-command
+call-out failed, and a `UUID` and `PARENT` of the device.
+
+```
+echo $config | $script -e "notify" -a "start" -s "success" -u "f0bb71ac-9b72-4d2d-bbdb-67f41d3cd26e" -p "matrix"
+```
+
+## Get Attributes Event
+
+Get events are invoked to the script as a "get" `EVENT` to request extraneous
+data from the mediated device in the case where mdevctl cannot easily acquire
+them (e.g. from a sysfs device, or an active mdev started by mdevctl). This
+event is always paired with the "attributes" `ACTION`.
+
+This call-out is made during the define and list commands. For define,
+this will acquire the device attributes when creating a device configuration
+by providing the UUID for an existing device that was not previously defined
+by mdevctl. When list is provided the `--dumpjson` option, this will acquire
+the device attributes when providing an active mdev that is queried from sysfs.
+
+A get call-out will invoke a script with the "get" `EVENT`, an "attributes"
+`ACTION`, a "none" `STATUS`, and a `UUID` and `PARENT` of the device.
+
+```
+echo $config | $script -e "get" -a "attributes" -s "none" -u "f0bb71ac-9b72-4d2d-bbdb-67f41d3cd26e" -p "matrix"
+```
+
+The expected output from the script is a JSON formatted array of device
+attributes that may be easily plugged into the device configuration:
+
+```
+[
+    {
+        "assign_adapter": "2"
+    },
+    {
+        "assign_domain": "0x3b"
+    }
+]
+```
+
+## Auto Start
+
+It is worth mentioning that for auto start devices, a pre/post call-out and
+notifier is invoked for each device. The parameters are the same as for a
+start command. The pre-command event is non-disruptive in this case as
+to allow mdevctl to attempt each device.
+
+All errors reported by the pre/post events are redirected to systemd.
+
+Note: if a notification script is used to convey information to another
+program or daemon, it is not guaranteed that the program will be started
+prior to mdevctl's invocation.
+
+## Script Installation
+
+For pre/post/get events, it may be the case that a script can be used to
+satisfy various device types. As such, a special "locator" script is required
+for mdevctl to find the appropriate script to execute. The locator script must
+accept a device type as a parameter and return via standard output either a
+valid path to the script or an empty string if the device type is not
+supported.
+
+For pre/post events, the locator scripts must reside within
+`/etc/mdevctl/callouts/command.d`, and for get events the locator scripts
+must reside within `/etc/mdevctl/callouts/get.d`. mdevctl will execute all
+locator scripts in the respective directory until either a valid path is
+returned or all scripts have returned an empty path.
+
+It is the responsibility of the call-out script to install its locator script
+to the appropriate directory.
+
+An example execution of a locator script:
+`/etc/mdevctl/callouts/command.d/ap_command.sh -t vfio_ap-passthrough`
+
+An example output from the locator script:
+`/usr/sbin/ap_check`
+
+Notifier event scripts must reside within
+`/etc/mdevctl/notification/notifiers.d/`. This event type does not require a
+locator script, and all scripts within this directory will be executed
+regardless of device type.
--- a/mdevctl
+++ b/mdevctl
@@ -609,69 +609,50 @@ case "$cmd" in
                 exit 1
             fi
 
-            write_config "$persist_base/$parent/$uuid"
-            if [ $? -ne 0 ]; then
-                exit 1
-            fi
+        else
 
-            $print_uuid
-            exit 0
-        fi
-
-        if [ -n "$uuid" ]; then
-            if [ -z "$parent" ]; then
-                if [ ! -L "$mdev_base/$uuid" ] || [ -n "$type" ]; then
-                    usage
+            if [ -n "$uuid" ]; then
+                if [ -z "$parent" ]; then
+                    if [ ! -L "$mdev_base/$uuid" ] || [ -n "$type" ]; then
+                        usage
+                    fi
+                    parent=$(basename $(realpath "$mdev_base/$uuid" | sed -s "s/\/$uuid//"))
+                    type=$(basename $(realpath "$mdev_base/$uuid/mdev_type"))
                 fi
 
-                parent=$(basename $(realpath "$mdev_base/$uuid" | sed -s "s/\/$uuid//"))
-                type=$(basename $(realpath "$mdev_base/$uuid/mdev_type"))
+                if [ -e "$persist_base/$parent/$uuid" ]; then
+                    echo "Device $uuid on $parent already defined, try modify?" >&2
+                    exit 1
+                fi
+            else
+                uuid=$(unique_uuid)
+                print_uuid="echo $uuid"
             fi
 
-            if [ -e "$persist_base/$parent/$uuid" ]; then
-                echo "Device $uuid on $parent already defined, try modify?" >&2
-                exit 1
+            if [ -z "$parent" ]; then
+                usage
             fi
-        else
-            uuid=$(unique_uuid)
-            print_uuid="echo $uuid"
-        fi
 
-        if [ -z "$parent" ]; then
-            usage
-        fi
+            if [ -z "$type" ]; then
+                usage
+            fi
 
-        if [ -z "$type" ]; then
-            usage
-        fi
+            if [ -n "$auto" ]; then
+                start="auto"
+            else
+                start="manual"
+            fi
 
-        if [ -n "$auto" ]; then
-            start="auto"
-        else
-            start="manual"
-        fi
+            set -o errexit
 
-        set -o errexit
-
-        mkdir -p "$persist_base/$parent"
-        set_config_key mdev_type "$type"
-        set_config_key start "$start"
-        write_config "$persist_base/$parent/$uuid"
-        if [ $? -eq 0 ]; then
-            $print_uuid
+            mkdir -p "$persist_base/$parent"
+            set_config_key mdev_type "$type"
+            set_config_key start "$start"
         fi
         ;;
     undefine)
         if [ -z "$uuid" ]; then
             usage
-        fi
-
-        set -o errexit
-
-        if [ -n "$parent" ]; then
-            rm -f "$persist_base/$parent/$uuid"
-        else
-            find "$persist_base" -name "$uuid" -type f | xargs rm -f
         fi
         ;;
     modify)
@@ -743,8 +724,6 @@ case "$cmd" in
 
             del_attr_index "$index"
         fi
-
-        write_config "$file"
         ;;
     start)
         set -o errexit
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
 ```
 
 See `mdevctl --help` or the manpage for more information.
+
```
cohuck, line 271: I'm wondering where most of that documentation should live. We want to have it easily accessible for someone wanting to write some scripts, who should not need to bother to look at the source code.

I guess that means we need to add something to the man page: where scripts should live, what they should look like, when they will be invoked, etc.

README.md
```diff
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+# Invoking External Scripts for Device Events
+
+Certain mediated devices may require additional operations or functionality,
+such as configuration checking or event reporting, before or after mdevctl
+executes a command. In order to remain device-type agnostic, mdevctl will
+"call-out" to external scripts to handle the extraneous work. These scripts
```
cohuck, line 277: s/extraneous/additional/ ?

README.md
```diff
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+`/etc/mdevctl/callouts/command.d`, and for get events the locator scripts
+must reside within `/etc/mdevctl/callouts/get.d`. mdevctl will execute all
+locator scripts in the respective directory until either a valid path is
+returned or all scripts have returned an empty path.
+
+It is the responsibility of the call-out script to install its locator script
+to the appropriate directory.
+
+An example execution of a locator script:
+`/etc/mdevctl/callouts/command.d/ap_command.sh -t vfio_ap-passthrough`
```
cohuck, line 471: Should we also ship a simple example script with mdevctl?

README.md
collinwalling: Right. I was wondering if a formal API documentation would be necessary here (I'm honestly not sure if all this information belongs in the README...)

As for a man page entry, perhaps the "Script Parameters" section with expanded description for each event type, action, state etc would suffice? I'd have to draft it up to see how it looks :)

README.md
collinwalling: I can check in with Matt and see what he thinks about including his vfio_ap-passthrough scripts. If not, I can at the very least provide my debugging script as a notifier (it simply echoes the call-out parameters to systemd).

README.md
collinwalling: Let me know what you think of https://gist.github.com/collinwalling/00b116888842f957f7f8e08e2e26c751

It has the man page formatting, so hopefully it's not too tricky to parse.

cohuck: Thanks, that is what I was thinking about. Seems sane at first glance, but I haven't done any proper review yet.
Maybe the readme should still contain a pointer to this feature, though?

README.md
cohuck: Yes, something simple like that debugging script. Device specifics might confuse people not familiar with that device type.

mdevctl
```diff
@@ -609,70 +609,51 @@ case "$cmd" in
...
+            if [ -z "$type" ]; then
+                usage
+            fi
 
-        if [ -n "$auto" ]; then
-            start="auto"
-        else
-            start="manual"
-        fi
+            if [ -n "$auto" ]; then
```
cohuck, line 842: Hm, this looks like you introduced some tabs (hard to see in the web interface)... we've been mostly tab-free so far, maybe stay in line with the existing code?

mdevctl
```diff
@@ -1015,3 +986,34 @@ case "$cmd" in
...
+    start)
+        start_mdev "$uuid" "$parent" "$type" "$print_uuid"
+        exit $?
+        ;;
+    stop)
+        set -o errexit
+
+        remove_mdev "$uuid"
+        ;;
+esac
```
cohuck, line 1248: This is a bit hard to review, but seems sane after a readthrough.
The whole program is getting a bit too complex for a shell script, though :/

cohuck: Note that the comment above applies to commit 604973a61af82e01f25ceeba788383414d8c97be, not 8b37cec8b6c6d67d8480b1eb13ebacbbb0801260, where it shows up via the web interface for me as well :(

mdevctl
collinwalling: I'm actually not sure what that dark-green bar means. I would agree with you that it looks like a tab, but when I look at my editor I don't see any tabs at all (I have "invisibles" turned on, so I just see a lot of dots indicating white space).

README.md
collinwalling: Good idea. I can keep a short description in the README explaining this capability and end it with a "see manpage for more information".

README.md
collinwalling: Sounds reasonable. How would you recommend I include this script? Should I just toss it in the repo as a separate commit?

cohuck: Adding it via a separate commit sounds good.

mdevctl
cohuck: OK, so more web interface weirdness... I'll do a proper review on a local checkout anyway.

mdevctl
cohuck: Just hold off for now; I still have to do a full review anyway.

mdevctl
```diff
@@ -987,6 +1143,13 @@ case "$cmd" in
         ;;
 esac
 
+get_device_parameters "$uuid" "$parent" "$type"
+
+callout "pre" "$cmd" "$uuid" "$parent" "$type"
```
cohuck, line 1219: Looking at the end result, I wonder whether we should rather split the commands into modifying and non-modifying and do the pre-/post-callouts only for the modifying commands.

Also, what is supposed to happen for commands not operating on a device, like 'version'?

mdevctl
```diff
@@ -1009,11 +1170,17 @@ case "$cmd" in
...
 esac
+if [ $? -eq 0 ]; then
+    state="success"
+else
+    state="failure"
+fi
+
+callout "post" "$cmd" "$uuid" "$parent" "$type"
+
+callout_notify "$cmd" "$uuid" "$parent" "$type"
```
cohuck, line 1257: Does the notify callback make sense for the general case (i.e. also for non-modifying commands)?

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+    if dir_is_empty "$callout_script"; then
+        return 0
+    fi
+
+    if ! get_callout_script "$callout_script"; then
+        return 0
+    fi
+
+    # Save content of $config to $conf variable
+    # Allow scripts to check the device configuration
```
cohuck, line 434: Nit: trailing whitespace

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+
+    # Save content of $config to $conf variable
+    # Allow scripts to check the device configuration 
+    # even after the device removed from the system
+    if [ "$conf" == "" ]; then
+        conf=$(dump_config)
+    fi
+
+    set -o pipefail
+
```
cohuck, line 441: Nit: whitespace in line

README.md
```diff
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+of mdevctl's primary command execution, or "none" if the pre-command
+call-out failed, and a `UUID` and `PARENT` of the device.
+
+```
+echo $config | $script -e "notify" -a "start" -s "success" -u "f0bb71ac-9b72-4d2d-bbdb-67f41d3cd26e" -p "matrix"
+```
+
+## Get Attributes Event
+
+Get events are invoked to the script as a "get" `EVENT` to request extraneous
```
cohuck, line 407: s/extraneous/additional/

README.md
```diff
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+data from the mediated device in the case where mdevctl cannot easily acquire
+them (e.g. from a sysfs device, or an active mdev started by mdevctl). This
+event is always paired with the "attributes" `ACTION`.
+
+This call-out is made during the define and list commands. For define,
+this will acquire the device attributes when creating a device configuration
+by providing the UUID for an existing device that was not previously defined
+by mdevctl. When list is provided the `--dumpjson` option, this will acquire
+the device attributes when providing an active mdev that is queried from sysfs.
```
cohuck, line 416: Does that mean that the get script is invoked for every device in the list, if you don't restrict it to a concrete device? The doc might be a bit more precise here.

mdevctl
collinwalling: Or perhaps the non-modifying commands should exit early?

The code does not support pre/post and notifer call-outs for version, list, and types. I'll make a note to add that to the docs.

mdevctl
collinwalling: I don't think so. Any information derived from the non-modifying commands can be captured directly from stdout if the user cares about it.

README.md
collinwalling: Yes, it is for every device in the list.

mdevctl
cohuck: Hm. Maybe it would be easier to first handle all non-modifying commands, then exit, and then handle the modifying commands in a prep -> pre-callout -> processing -> post-callout sequence? If we have too many early exits, the code becomes too complicated to follow.

README.md
cohuck: Maybe add "Note that this will happen for each device in the list."?

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
 
+callout_script="$persist_base/callouts/command.d"
+callout_notification="$persist_base/notification/notifiers.d"
+
+state="none"
+
+dir_is_empty() {
+    path="$1"
+
+    if [ -n "$(ls -A "$path")" ]; then
```
awilliam, line 343: Why are we including other hidden files here?  We won't consider them for callout scripts per get_callout_script() below

README.md
cohuck: BTW, do you have any scripts to peek at available already? That might help with reviewing the code and the documentation.

collinwalling: Sure. This is just a basic notifier script as mentioned before: https://gist.github.com/collinwalling/88c06b6803811c2678eb06b4ddaee90f

README.md
collinwalling: I can change the last sentence to something like:

> ... this will acquire the device attributes **_for each active mdev_** that is queried from sysfs.

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+
+    return 0
+}
+
+callout() {
+    event="$1"
+    action="$2"
+    uuid="$3"
+    parent="$4"
+    type="$5"
```
awilliam, line 417: Unused?

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+    (sed "s/^/$(basename $(realpath "$script")): /" >&2)
+
+    return $?
+}
+
+callout_notify() {
+    action="$1"
+    uuid="$2"
+    parent="$3"
+    type="$4"
```
awilliam, line 481: $type is unused again?

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+    # even after the device removed from the system
+    if [ "$conf" == "" ]; then
+        conf=$(dump_config)
+    fi
+
+    set -o pipefail
+
+    for n in $callout_notification/*; do
+        echo "$conf" | $n -e "notify" -a "$action" -s "$state" -u "$uuid" -p "$parent" 2>&1 |
+        (sed "s/^/$(basename $(realpath "$n")): /")
```
awilliam, line 504: Here as above, but not redirected to /dev/null, so I can see how something reaches the final pipe.

mdevctl
```diff
@@ -423,10 +570,19 @@ case ${1} in
                 fi
 
                 if [ "$(get_config_key start)" == "auto" ]; then
-                    start_mdev "$uuid" "$parent" "$(get_config_key mdev_type)"
-                    if [ $? -ne 0 ]; then
-                        echo "Failed to create mdev $uuid, type $(get_config_key mdev_type) on $parent" >&2
-                        # continue...
+                    callout "pre" "start" "$uuid" "$parent" "$(get_config_key mdev_type)" 2>&1 | systemd-cat -t $(basename $0) -p err
```
awilliam, line 619: Is this not imposing systemd-cat as an implicit notification provider?

mdevctl
```diff
@@ -316,8 +316,26 @@ unique_uuid() {
...
+    parent="$2"
+    type="$3"
+
+    set_config_key mdev_type "$type"
+    set_config_key start 'manual'
+
+    callout_get "get" "attributes" "$uuid" "$parent" "$type"
+    if [ $? -eq 0 ]; then
+        if [ "$get_attr" != [{}] ]; then
+            attrs=$get_attr
```
awilliam, line 335: We overwrite any other attrs in favor of what this script provided??

mdevctl
```diff
@@ -428,6 +445,35 @@ callout() {
     return $?
 }
 
+callout_get() {
+    local event="$1"
+    local action="$2"
+    local uuid="$3"
+    local parent="$4"
+    local type="$5"
```
awilliam, line 453: $type unused?  Oh, are these getting implicitly used by get_callout_script?  Too subtle for me.

mdevctl
```diff
@@ -1048,6 +1105,20 @@ case "$cmd" in
...
+                if [ -n "$dumpjson" ]; then
+                    callout_get "get" "attributes" "$u" "$p" "$type"
+                    if [ $? -eq 0 ]; then
+                        if [ "$get_attr" != [] ]; then # callout_get script has been invoked
+                            if [ "$get_attr" == [{}] ]; then
+                                get_attr=[]
+                            fi
+
+                            json_tmp=$(echo "$json_tmp" | jq --argjson att "{\"attrs\":$get_attr}" \
+                                    '(.matrix | .[] | .[] | .) += $att')
```
awilliam, line 1117: .matrix??  Looks like this is not generic.

README.md
```diff
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+
+A get call-out will invoke a script with the "get" `EVENT`, an "attributes"
+`ACTION`, a "none" `STATUS`, and a `UUID` and `PARENT` of the device.
+
```
awilliam, line 420: Why do we need a separate script for "get" events?

README.md
```diff
@@ -268,3 +268,212 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+It is the responsibility of the call-out script to install its locator script
+to the appropriate directory.
+
+An example execution of a locator script:
+`/etc/mdevctl/callouts/command.d/ap_command.sh -t vfio_ap-passthrough`
+
+An example output from the locator script:
+`/usr/sbin/ap_check`
+
```
awilliam, line 475: I think we ant to make sure parameters for the "locator" script and the script itself don't collide, such that they could be one in the same. ie. the script might return $0 to indicate itself, right?

collinwalling: I think the idea is to have a directory that (ideally) contains only locator scripts. mdevctl will iterate through each file in this directory by invoking each locator script with `-t device_type` until one of them returns a path to the actual script.

Does it make sense to store the actual scripts in this directory? It might be confusing if it contains a mix of locators & actuals.

README.md
collinwalling: If I recall from our discussion we had a few weeks back, it made sense to have a separate directory for each call-out event (one for "get", one for "pre/post" aka "commands", and one for "notify"). The locator scripts in get.d and command.d may point to the same actual script.

mdevctl
zhuoyingcai: I'll add a condition to make sure attrs don't get overwrite.

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+    # Allow scripts to check the device configuration 
+    # even after the device removed from the system
+    if [ "$conf" == "" ]; then
+        conf=$(dump_config)
+    fi
+
+    set -o pipefail
+    
+    echo "$conf" | $script -e "$event" -a "$action" -s "$state" -u "$uuid" -p "$parent" 2>&1 > /dev/null |
+    (sed "s/^/$(basename $(realpath "$script")): /" >&2)
```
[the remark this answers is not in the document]
awilliam: Aha, thanks.  But why do we need the parens around the sed command?  Testing similar syntax, I can't tell that it does anything.

README.md
awilliam: Hmm, the "notify" I understand, especially given your demo that trivially enabled debugging.  We can also expect that different entities might want to drop in new notify scripts and they have different conventions of calling them all vs finding a single match.  I can't think of similar reasons for separate "get" scripts.

README.md
awilliam: This "locator" concept is a little foreign to me, maybe we discussed it, maybe I even proposed it :-P, I'm not sure.  I sort of assumed that we'd probe the scripts in the directory until one returned success and just keep using it rather than having it return a path to the real script.  The latter seems a little tricky as that path could be different per distro or method of install.  If we just call that script again after probing, then the thing in the scripts.d directory could be a link to the real script or it could the script itself, I'm not sure we care.  If we assume a single script handles get/pre/post then we could just add a test/query/nop event where the script returns pass/fail if it recognizes the type.  Or maybe we expect scripts to return values telling us whether to continue running scripts or stop.  For instance, we could always call scripts until one tells us to stop so that they could be named to impose an ordering, ie. 00_foo vs 99_bar.  I know we already have performance issues, but I'm not sure we should let the current bash implementation decide how we lay out our script support.

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+        # $config was not assigned before initial callout and config file exists
+        # We need to read the config file to assign $config and $attrs
+            read_config "$persist_base/$parent/$uuid"
+        fi
+    fi
+
+    if [ -z "$type" ]; then
+        type="$(get_config_key mdev_type)"
+    fi
+
```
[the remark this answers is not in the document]
cohuck: I think this function might benefit from either restructuring it a bit, or from adding more comments what we're looking for.

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+# Not all commands provide all necessary info for the callouts 
+# Some sysfs devices need some extra work
+get_device_parameters() {
+    uuid="$1"
+    parent="$2"
+    type="$3"
+
+    if [ -z "$parent" ]; then
+        count=$(find "$persist_base" -name "$uuid" -type f | wc -l)
```
cohuck, line 375: This 'count' variable is really confusing; in the end, we are checking for existence or non-exisitence, and ambiguity would be an error?

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+            fi
+        else
+            return 1
+        fi
+    fi
+    
+    if [ $(echo "$config" | jq 'length') -eq 0 ]; then 
+        # Start and stop a device without defined config file
+        # $config was not assigned before initial callout and config file does not exists
+        # We need to assigned $config with parameters passed by the user
```
cohuck, line 394: Maybe
```
# We arrive here when starting or stopping a device that does 
# not have a defined config file.
# In that case, $config has not yet been populated before the initial
# callout, and we need to populate it with the parameters that were
# passed by the user.
```

mdevctl
```diff
@@ -314,6 +314,153 @@ unique_uuid() {
...
+        if [ ! -e "$persist_base/$parent/$uuid" ]; then
+            set_config_key mdev_type "$type"
+            set_config_key start 'manual'
+        else
+        # Undefine and stop a device
+        # $config was not assigned before initial callout and config file exists
+        # We need to read the config file to assign $config and $attrs
```
cohuck, line 400: What about
```
# We arrive here when undefining and stopping a device.
# Again, $config has not yet been populated before the 
# initial callout, but we can simply read the existing
# config file to populate it.
```

mdevctl
cohuck: We currently carry systemd only as a BuildRequires. Do we want to change that?

mdevctl
cohuck: Weirdly, the green bar seems to be gone now... I don't even pretend to understand this, but the local checkout looks fine.

README.md
cohuck: My proposal would be:
* all callouts have to live in /etc/mdevctl/callouts/scripts.d/
* callout scripts are called with 'type' 'callout' 'callout specifc parameters'
  * A script first checks if it can handle 'type' and will exit with a 'not me' exit code if not
  * The script can invoke whatever scripts, binaries, ... for individual callouts, or handle them itself
  * It should exit with an exit code different from 'not me' if something went wrong, and 0 if everything is fine, or it doesn't do anything for that specific callout
  * If there are scripts that should be invoked for specfic device types, and scripts that are more of a catch-all for types that don't need more specific processing, the specialized script should be e.g. 20_specificfoo and the generic one 60_foo
* mdevctl will call one script after the other, continuing for a 'not me' exit code, assuming success for a zero exit code, and failure in all other cases
  * It might want to cache the first successful invocation.
 
I'm wondering how much we should optimize for many scripts. Do we expect scripts to come with a distro mdevctl package (i.e. all scripts will be installed), or that there will be per-device type or per-architecture script packages (i.e. we'll see vfio-ap related scripts only on an s390 system)?

mdevctl
zhuoyingcai: I will add conditions to make sure that attrs don't get overwritten.

mdevctl
cohuck: How does that work for 'undefine'? We allow ambiguity there (configs for a uuid are removed for all parents).

zhuoyingcai: In Bash, all variables by default are defined as global, even if declared inside the function. $config is also a global variable declared at line 20. Once $uuid, $config, $parent and $type variables are populated, they are still accessible even if the config file is removed.

Before the initial callout of 'undefine', $uuid is provided by the user. Line 356 to 371 is going to populate $parent based on the $uuid. If no error occurs, line 384 will populate $config by reading into the existing config file of the $uuid and $parent. Lastly, line 389 is going retrieve the $type.

README.md
collinwalling: Okay. I think this simplifies things a bit. We can drop the "locator" script concept and have the "actual" scripts determine whether or not they support the device type. 

A few questions:
1. As for the prefixed numbering scheme, how should that be determined and who should set those values? The values 20_ for specific and 60_ for generic seems sane to me... does it simply need to be documented?
2. Should mdevctl execute the generic script(s) after successfully executing the specific (singular) script?

README.md
collinwalling: Hmmm, okay. If neither of us can think of a good reason for separating the two callouts, then let's move forward with merging them :)

mdevctl
cohuck: Yes, but don't we fail to populate parent if we have multiple configs?

README.md
cohuck: Regarding numbering, I think we can advise some suggested numbers, but I'm not sure we want to mandate them.

I don't think we want to invoke the generic script unconditionally. Maybe a script wants to first do the things that the generic script wants to do as well, and only afterwards do some specific things. If at most one script is executed (excluding checking), we are more flexible; the scripts can still put generic routines into helper scripts or so.

mdevctl
zhuoyingcai: That's correct. get_device_parameter function will return 1 if we have multiple configs. $parent is not populated, which means it's an empty string. Later, callout script from the initial callout should detect the invalid $parent, and output any error.

cohuck: But what does that mean for undefine, where we do allow ambuguity? We don't want to return an error in that case, and I don't see the callouts invoked for each occurence. What am I missing? (Also, isn't the way callout() evaluates positional parameters a problem?)

zhuoyingcai: Okay, I understand what you mean now. Sorry that I didn't pay attention to the ambiguity of undefine.

I took some time and tested undefine multiple configs with the same uuid. I realized that the callout script didn’t invoke at all for this situation.

The get_device_parameters function returns 1 when it detects there are multiple configs exist. Therefore, $type variable is not populated. Then the callout function returns 0 because the get_callout_script function didn’t locate any script if no type was provided. In the end, all configs are still being removed, but without invoking any callouts.

We had some discussion on this topic. I think undefine is the only case that allows ambiguity (correct me if I'm wrong). What we plan to do for undefine if multiple configs have the same uuid is to invoke callouts for an individual device, and remove one device at a time. Exit the program until all the configs with the same uuid are checked.

Please let me know if you have any suggestions for this. Thank you!

cohuck: Invoking the callouts for any of the uuid/parent combinations looks like the right thing to do.

As for implementation, I'm not sure. You'd probably need to loop both before and after removing the definitions, but any caching of values would probably be ugly.

mdevctl
```diff
@@ -1009,11 +1170,17 @@ case "$cmd" in
         ;;
     start)
         start_mdev "$uuid" "$parent" "$type" "$print_uuid"
-        exit $?
```
collinwalling: Should we make sure we capture the return code here? I'm not sure if any external scripts that invoke mdevctl need to listen for a proper return code when attempting to start a device.

Additionally, should mdevctl `exit 0` when status is "success" and `exit 1` when status is "failure" (after the post callout and notify, of course)?

mdevctl
collinwalling: I'm sorry for letting this one linger for so long. As Joy mentions, systemd-cat allows mdevctl to report any issues from the callouts during the auto-start procedure. Since this gets fired off by a udev rule during system boot, we need to print this stuff somewhere that isn't stderr so the user can look back on it (e.g. via journalctl).

I'm not familiar with BuildRequires, but from what I was able to gather: this is some meta data associated with a package? Is this something we need to modify / request on our end or is that ultimately up to the distro folks?

cohuck: Hardcoding systemd-cat here would mean adding a hard dependency on systemd. We currently only depend on systemd when building the rpm (see 3e10c587173c1 for an explanation), but not for operation.
So, I think we have two options here:
- add a dependency on systemd, and use systemd-cat
- use something different from systemd-cat to provide a similar function, maybe make it configurable

mdevctl
cohuck: I think a non-zero return code on failure is a good idea regardless whether anybody is relying on it right now.

mdevctl
awilliam: Is logger(1) from util-linux package meant to be a more generic way to write to system logs rather than using systemd specific tools?

collinwalling: Good find. Functionally, logger works perfectly. @cohuck if this aligns with your second bullet-point, then I say lets go for it.

cohuck: Yes, logger is probably a good way to implement this.

collinwalling:
Thanks for the feedback. v2 can be found at #35

PR #33, opened 2021-05-04
[tests] Add system-level roundtrip test

smitterl:
Add a test script to cover critical scenarios at least once.

The test script requires the mdev test device mtty. This device
was selected because it is available for all kernels. Details
Documentation/driver-api/vfio-mediated-device.rst

However, as the test device doesn't have attributes
'modify' is confirmed to at least try to set the values.

This test script is not thought to be an extensive test suite.
The intention is to rather quickly provide a minimal test suite
to assist code refactor necessary to add unit tests in a future patch.

Signed-off-by: Sebastian Mitterle <smitterl@redhat.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ mdevctl.spec
 *.tar.gz
 *.rpm
 */
+!tests/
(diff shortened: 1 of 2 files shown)


cohuck:
Is this still of interest with the rust rewite, or do the tests included there cover this scenario?

smitterl:
Nope, closing :)

PR #34, opened 2021-05-04
[safe-refactor] refactor code to allow for easier unit testing

smitterl:
Existing behavior is not changed.

As can be seen, there is no change in logic, only parts of the code have
been moved into smaller functions in order to allow for unit testing them
more easily.

As unit tests will want to source the script file in order to test the
functions in isolation, a switch has been added to avoid script execution
when the script is sourced.

1. Define top-level functions
 - `parse_args` to hold the argument parsing logic. After this function
   is run, global arguments have been assigned to script global variables
   to be used in other functions.
 - `run_command` contains the command selection and delegation to handling
   functions.
2. Define handling functions
 - Each case' commands have been moved from `run_command` to top-level
   functions. This way a unit test can call for example only the function
   under test.
 - A unit test could look like:
   ```bash
    #!/bin/bash
    source mdevctl
    source setup # some logic to use local $parent_base and $mdev_base
                 # for testing instead of the real ones
    res=0
    parent=some_parent
    type=some_type
    result=$(define_device)
    if ! [[ "$?" -eq "0" ]] || [[ -z "$result" ]]; then
        echo "Failed to define device."
        res=1
    fi
    exit res
   ```
3. Add switch to avoid execution during sourcing.

Signed-off-by: Sebastian Mitterle <smitterl@redhat.com>

### DIFF
--- a/mdevctl
+++ b/mdevctl
@@ -314,6 +314,376 @@ unique_uuid() {
     echo "$uuid"
 }
 
+
+define_device() {
+    if [ -n "$jsonfile" ]; then
+        if [ ! -r "$jsonfile" ]; then
+            echo "Unable to read file $jsonfile" >&2
+            exit 1
+        fi
+        if [ -n "$type" ]; then
+            echo "Device type cannot be specified separately from $jsonfile" >&2
+            exit 1
+        fi
+        if [ -z "$parent" ]; then
+            echo "Parent device required to define device via $jsonfile" >&2
+        fi
+        if [ -z "$uuid" ]; then
+            uuid=$(unique_uuid)
+            print_uuid="echo $uuid"
+        fi
+        if [ -e "$persist_base/$parent/$uuid" ]; then
+            echo "Cowardly refusing to overwrite existing config for $parent/$uuid" >&2
+            exit 1
+        fi
+        set -o errexit
+        read_config "$jsonfile"
+        if [ $? -ne 0 ]; then
+            echo "Error reading $jsonfile" >&2
+            exit 1
+        fi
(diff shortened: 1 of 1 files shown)


smitterl:
The code was tested dynamically with the system-level test script from https://github.com/mdevctl/mdevctl/pull/33
```bash
# make rpm
# rpm -i noarch/mdevctl-0.80-1.el8.noarch.rpm
# /root/roundtrip_test
- Confirm that test device is a known type
- Define and list device without uuid on mtty-1
- Define and list created device on mtty-2
- Stop known device on mtty-2
- Modify known device on mtty-2 and check json
- Start known device tries to set attribute but fails on non-existing attribute
- Delete non-existing attribute on known device
- Define device from json, change start type
- Start known devices
- Stop all known devices
- Undefine all known devices
```

smitterl:
Unfortunately, the diff looks rather ugly and obfuscates the little change that has happened. It might be easier to compare the actual file manually - https://github.com/mdevctl/mdevctl/blob/475e1f2cc583fa571e23b3b6426a9cd481344cb3/mdevctl

awilliam:
Thanks for the pull requests!  This looks useful.  We're currently considering a rebase to a rust based rewrite of mdevctl, where a (nearly) functionally complete prototype can be found here: https://gitlab.com/jjongsma/mdevctl-experimental  As you'll see there, Jonathon has already include several tests.  It would be great if you could review how your testing matches to this new version.  Thanks!

cohuck:
I think this has been obsoleted by the rust rewrite; is there still something we want to do on top of that?

smitterl:
Yep, closing. :)

PR #35, opened 2021-05-04
[v2] Invoking External Scripts for Device Events

collinwalling:
Version 2 of the call-out implementation. Previous version is #32.

We are aware of the mdevctl-experimental project, but would still like to propose this as a core functionality of mdevctl (present and future) to allow for device-specific configurations.

Changelog:
 - moved the bulk of the documentation from README to the man page
 - included a basic logger notifier script as an example of the call-out functionality
 - restructured commits
    - call-outs for define, modify, and stop are rather straight-forward and are supported via a single patch
    - call-outs for manual/auto start is a patch on its own to underline the auto-start special case
    - call-outs for undefine is also a patch on its own since it needs to be handled differently than the other commands due to handling multiple devices at once
       - Handle undefine in the first command block
       - Invoke callout scripts before and after removing each config file.
 - "locator script" concept is dropped
    - all call-out scripts for pre/post/get events are now stored in /etc/mdevctl.d/callouts/scripts.d/
 - The new callout script API format:  `echo "$conf" | $script -t "$type" -e "$event" -a "$action" -s "$state" -u "$uuid" -p "$parent"`
    - pre/post/get now passes device type as the first parameter, this should be the first thing the script checks to confirm if it supports the  device type, returning EOPNOTSUPP (95) otherwise.
    - notify is unchanged
 - replaced systemd-cat with logger utility
 - exit code from mdevctl now reflects SUCCESS/FAILURE state for define, start, stop, and modify
 - various code cleanups

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ PREFIX=/usr
 UDEVDIR=$(shell pkg-config --variable=udevdir udev)
 SBINDIR=$(PREFIX)/sbin
 CONFDIR=/etc/mdevctl.d
+CALLOUT_CMD_DIR=$(CONFDIR)/callouts/scripts.d
+CALLOUT_NOTIFIER_DIR=$(CONFDIR)/notification/notifiers.d
 MANDIR=$(PREFIX)/share/man
 NAME=mdevctl
 MDEVCTL_VER=$(shell ./mdevctl version)
@@ -44,6 +46,8 @@ install:
 	mkdir -p $(DESTDIR)$(MANDIR)/man8
 	install -m 644 mdevctl.8 $(DESTDIR)$(MANDIR)/man8/
 	ln -sf mdevctl.8  $(DESTDIR)$(MANDIR)/man8/lsmdev.8
+	mkdir -p $(DESTDIR)$(CALLOUT_CMD_DIR)
+	mkdir -p $(DESTDIR)$(CALLOUT_NOTIFIER_DIR)
 
 clean:
 	rm -f mdevctl.spec *.src.rpm noarch/*.rpm *.tar.gz
--- a/README.md
+++ b/README.md
@@ -267,4 +267,28 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
     @{2}: {"assign_domain":"0xff"}
 ```
 
+# Invoking External Scripts for Device Events
+
+Certain mediated devices may require additional operations or functionality,
+such as configuration checking or event reporting, before or after mdevctl
+executes a command. In order to remain device-type agnostic, mdevctl will
+"call-out" to external scripts to handle the extraneous work. These scripts
+are associated with a specific device type to perform any operations or
+additional functionality not handled within mdevctl. Additionally, external
+programs may wish to receive notifications of any action performed by mdevctl
+to e.g. respond to any device changes or keep device management in parallel.
+
+A call-out script is invoked at various points during an mdevctl command
+process and are categorized by an "event" paired with an "action", along
+with information regarding the mediated device. Two main event types are
+"pre" and "post", which are invoked before and after primary command
+execution respectively. The same call-out script is invoked for both events.
+
+A "notify" event script is invoked to report the status of mdevctl's
+command results. This may be used to signal external programs of changes
+made to a mediated device, or simply to assist with debugging efforts.
+
+A "get" event script is invoked for the define and list commands to acquire
+device attributes from sysfs.
+
 See `mdevctl --help` or the manpage for more information.
--- a/logger.sh
+++ b/logger.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/bash
+# A basic utility script used for debugging notification call-out events
+# in mdevctl. Output can be observed via system logs.
+
+#stdin | -e event -a action -s state -u uuid -p parent
+shift
+event=$1
+shift 2
+action=$1
+shift 2
+state=$1
+shift 2
+uuid=$1
+shift 2
+parent=$1
+json=$(</dev/stdin)
+
+if [[ "$event" == "notify" ]]; then
+	echo "Event: $event, Action: $action, State: $state, UUID: $uuid, Parent: $parent, JSON:" | logger -t "mdevctl_logger"
+	echo "$json" | logger -t "mdevctl_logger"
+fi
--- a/mdevctl
+++ b/mdevctl
@@ -823,12 +823,48 @@ case "$cmd" in
             usage
         fi
 
-        set -o errexit
-
         if [ -n "$parent" ]; then
-            rm -f "$persist_base/$parent/$uuid"
+            read_config "$persist_base/$parent/$uuid"
+            type="$(get_config_key mdev_type)"
+
+            callout "pre" "$cmd" "$uuid" "$parent" "$type"
+            if [ $? -eq 0 ]; then
+                rm -f "$persist_base/$parent/$uuid"
+                if [ $? -eq 0 ]; then
+                    state="success"
+                else
+                    state="failure"
+                fi
+
+                callout "post" "$cmd" "$uuid" "$parent" "$type"
+            fi
+
+            callout_notify "$cmd" "$uuid" "$parent" "$type"
         else
-            find "$persist_base" -name "$uuid" -type f | xargs rm -f
+            undef_configs=$(find "$persist_base" -name "$uuid" -type f)
+
+            if [ -n "$undef_configs" ]; then
+                for p in $undef_configs; do
+                    parent=$(basename $(realpath "$p" | sed -s "s/\/$uuid//"))
+                    read_config "$persist_base/$parent/$uuid"
+                    type="$(get_config_key mdev_type)"
+
+                    callout "pre" "$cmd" "$uuid" "$parent" "$type"
+                    if [ $? -eq 0 ]; then
+                        rm -f "$persist_base/$parent/$uuid"
+                        if [ $? -eq 0 ]; then
+                            state="success"
+                        else
+                            state="failure"
+                        fi
+
+                        callout "post" "$cmd" "$uuid" "$parent" "$type"
+                    fi
+
+                    callout_notify "$cmd" "$uuid" "$parent" "$type"
+                    script=""
+                done
+            fi
         fi
 
         exit 0
--- a/mdevctl
+++ b/mdevctl
@@ -580,11 +580,22 @@ case ${1} in
                 fi
 
                 if [ "$(get_config_key start)" == "auto" ]; then
-                    start_mdev "$uuid" "$parent" "$(get_config_key mdev_type)"
-                    if [ $? -ne 0 ]; then
-                        echo "Failed to create mdev $uuid, type $(get_config_key mdev_type) on $parent" >&2
-                        # continue...
+                    callout "pre" "start" "$uuid" "$parent" "$(get_config_key mdev_type)" 2>&1 | logger -t $(basename $0) -p err
+                    if [ $? -eq 0 ]; then
+                        start_mdev "$uuid" "$parent" "$(get_config_key mdev_type)"
+                        if [ $? -ne 0 ]; then
+                            echo "Failed to create mdev $uuid, type $(get_config_key mdev_type) on $parent" >&2
+                            # continue...
+
+                            state="failure"
+                        else
+                            state="success"
+                        fi
+
+                        callout "post" "start" "$uuid" "$parent" "$(get_config_key mdev_type)" 2>&1 | logger -t $(basename $0) -p err
                     fi
+
+                    callout_notify "start" "$uuid" "$parent" "$type" | logger -t $(basename $0) -p notice
                 fi
             fi
         done
@@ -978,9 +989,6 @@ case "$cmd" in
                 print_uuid="echo $uuid"
             fi
         fi
-
-        start_mdev "$uuid" "$parent" "$type" "$print_uuid"
-        exit $?
         ;;
     stop)
         if [ -z "$uuid" ]; then
@@ -1174,6 +1182,9 @@ case "$cmd" in
         ;;
     modify)
         write_config "$file"
+        ;;
+    start)
+        start_mdev "$uuid" "$parent" "$type" "$print_uuid"
         ;;
     stop)
         remove_mdev "$uuid"
--- a/mdevctl
+++ b/mdevctl
@@ -616,7 +616,6 @@ case "$cmd" in
                     if [ ! -L "$mdev_base/$uuid" ] || [ -n "$type" ]; then
                         usage
                     fi
-
                     parent=$(basename $(realpath "$mdev_base/$uuid" | sed -s "s/\/$uuid//"))
                     type=$(basename $(realpath "$mdev_base/$uuid/mdev_type"))
                 fi
@@ -650,11 +649,6 @@ case "$cmd" in
             set_config_key mdev_type "$type"
             set_config_key start "$start"
         fi
-
-        write_config "$persist_base/$parent/$uuid"
-        if [ $? -eq 0 ]; then
-            $print_uuid
-        fi
         ;;
     undefine)
         if [ -z "$uuid" ]; then
@@ -668,6 +662,8 @@ case "$cmd" in
         else
             find "$persist_base" -name "$uuid" -type f | xargs rm -f
         fi
+
+        exit 0
         ;;
     modify)
         if [ -z "$uuid" ]; then
@@ -738,8 +734,6 @@ case "$cmd" in
 
             del_attr_index "$index"
         fi
-
-        write_config "$file"
         ;;
     start)
         set -o errexit
@@ -835,10 +829,6 @@ case "$cmd" in
         if [ -z "$uuid" ]; then
             usage
         fi
-
-        set -o errexit
-
-        remove_mdev "$uuid"
         ;;
     list)
         json="[]"
@@ -1007,5 +997,22 @@ case "$cmd" in
         else
             echo -en "$txt"
         fi
+        ;;
+esac
+
+case "$cmd" in
+    define)
+        write_config "$persist_base/$parent/$uuid"
+        if [ $? -eq 0 ]; then
+            $print_uuid
+        fi
+        ;;
+    modify)
+        write_config "$file"
+        ;;
+    stop)
+        set -o errexit
+
+        remove_mdev "$uuid"
         ;;
 esac
--- a/mdevctl
+++ b/mdevctl
@@ -314,6 +314,163 @@ unique_uuid() {
     echo "$uuid"
 }
 
+callout_script="$persist_base/callouts/scripts.d"
+callout_notification="$persist_base/notification/notifiers.d"
+
+state="none"
+
+dir_is_empty() {
+    path="$1"
+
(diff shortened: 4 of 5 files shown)


awilliam:
> We are aware of the mdevctl-experimental project, but would still like to propose this as a core functionality of mdevctl (present and future) to allow for device-specific configurations.

This doesn't make a lot of sense to me.  If we're on the cusp of deprecating the bash-based implementation, I think we'd want new functionality implemented in the new code base first, or at least concurrently.  Adding new functionality to the existing code base, without an equivalent for the new code base only introduces another barrier into the transition process.

collinwalling:
That's a fair response to my poorly-worded acknowledgement.

What I meant to state was that our intention with moving forward with this pull request was to ensure that callouts will become a core feature of mdevctl and would thus be considered as a requirement for a "feature complete" version of mdevctl-experimental. I think we should discuss if that means this code gets included in the bash version and used as a basis for callouts in the rust version, or something else entirely different. This is a highly-valuable feature for vfio_ap-passthrough mdevs.

Now, by all means this isn't a "we did the work, now someone else can do it" stance. At minimum I think it's a good idea to get involved with reviewing the rust version since it has gained a lot of traction and while it's still in the early stages (though I'd have to learn the language myself). I'll confer with Matt and Joy on this and how we should move forward.

cohuck:
Your bash implementation crossing with the Rust rewrite is unfortunate. I agree that this is an important feature, and that it should not be dropped.
Can you evaluate doing this work on top of the Rust version? If we get into agreement about the interfaces used by the external scripts, adding the same functionality to the rewrite should not be too bad.
I'll go ahead and review the interfaces/documentation here.

Makefile
```diff
@@ -2,6 +2,8 @@ PREFIX=/usr
 UDEVDIR=$(shell pkg-config --variable=udevdir udev)
 SBINDIR=$(PREFIX)/sbin
 CONFDIR=/etc/mdevctl.d
+CALLOUT_CMD_DIR=$(CONFDIR)/callouts/scripts.d
+CALLOUT_NOTIFIER_DIR=$(CONFDIR)/notification/notifiers.d
```
cohuck, line 6: I'm not sure whether we really should split callouts and notifications, they seem similar enough.
Will continue to read through the proposed interface and then come back here.

README.md
```diff
@@ -267,4 +267,28 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
     @{2}: {"assign_domain":"0xff"}
 ```
 
+# Invoking External Scripts for Device Events
+
+Certain mediated devices may require additional operations or functionality,
+such as configuration checking or event reporting, before or after mdevctl
+executes a command. In order to remain device-type agnostic, mdevctl will
+"call-out" to external scripts to handle the extraneous work. These scripts
```
cohuck, line 275: Still 'extraneous'... should be 'additional'.

logger.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/bash
+# A basic utility script used for debugging notification call-out events
+# in mdevctl. Output can be observed via system logs.
```
cohuck, line 3: Maybe add 'To use, drop this into ..."?

README.md
```diff
@@ -267,4 +267,28 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+"pre" and "post", which are invoked before and after primary command
+execution respectively. The same call-out script is invoked for both events.
+
+A "notify" event script is invoked to report the status of mdevctl's
+command results. This may be used to signal external programs of changes
+made to a mediated device, or simply to assist with debugging efforts.
+
+A "get" event script is invoked for the define and list commands to acquire
+device attributes from sysfs.
+
```
cohuck, line 293: Maybe refer to logger.sh as an example?

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+command success/failure (or "none" if the command was not executed), and the
+device's "TYPE", "UUID", "PARENT". The device's JSON configuration is provided
+via standard input.
+
+These scripts are invoked before and after an mdevctl's "primary command
+execution" (e.g. writing the device configuration file during define, or
+activating a device during start).
+
+All errors are redirected to standard error (except for auto-start call-out
+errors, which are reported to systemd).
```
cohuck, line 407: s/systemd/the system log/ ?

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+.IP "2. invoke pre-command call-out"
+.IP "3. primary command execution*"
+.IP "4. invoke post-command call-out*"
+.IP "5. invoke notifier"
+.IP "* step is skipped if 2 fails."
+.RE
+
+.SS CALL-OUT EVENTS
+Call-out event scripts are invoked with the following parameters below. For
+"pre", "post", and "get" call-outs, the "TYPE" parameter must be checked
```
cohuck, line 422: must be checked by the script?

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+to ensure the device is supported by the script and return an error otherwise
```
cohuck, line 423: "and an error must be returned otherwise"

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+.PP
+\fBPre-Command: \fR
+<\fICONFIG\fR> | SCRIPT \fR
+<\fB-t=\fR\fITYPE\fR \fB-e=\fR\fI"pre"\fR \fB-a=\fR\fIACTION\fR
+\fB-s=\fR\fI"none"\fR \fB-u=\fR\fIUUID\fR \fB-p=\fR\fIPARENT\fR>
+.RS 4
+A pre-command call-out is invoked once prior to primary command execution.
+This event will be paired with an action reflecting an mdevctl command and
+a status of "none". A non-zero return code will cause mdevctl to fire off
+a notification event (see below) and exit early. This event is not supported
```
cohuck, line 435: Maybe
"If the call-out script returns with a non-zero exit code, mdevctl will not perform the primary command execution, but will generate a notification event (see below) before exiting."

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+\fBPost-Command: \fR
+<\fICONFIG\fR> | SCRIPT \fR
+<\fB-t=\fR\fITYPE\fR \fB-e=\fR\fI"post"\fR \fB-a=\fR\fIACTION\fR
+\fB-s=\fR\fI"success"\fR|\fI"failure"\fR \fB-u=\fR\fIUUID\fR
+\fB-p=\fR\fIPARENT\fR>
+.RS 4
+A post-command call-out is invoked once after primary command execution.
+The action will reflect an mdevctl command. This event will be paired with
+a status of "success" if mdevctl returned a 0 during a primary command
+execution subroutine, or "failure" otherwise. A non-zero return code is
```
cohuck, line 449: Better "...if mdevctl was able to finish primary command execution successfully..." ?

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+.RE
+
+.PP
+\fBAuto-start: \fR
+<\fICONFIG\fR> | SCRIPT \fR
+<\fB-t=\fR\fITYPE\fR \fB-e=\fR\fI"pre"\fR|\fI"post"\fR\ \fB-a=\fR\fI"start"\fR
+\fB-s=\fR\fI"success"\fR|\fI"failure"\fR \fB-u=\fR\fIUUID\fR
+\fB-p=\fR\fIPARENT\fR>
+.RS 4
+For auto-start devices, a pre/post-command call-out is made for each device.
```
cohuck, line 461: "During processing of autostart devices, mdevctl will invoke the pre- and post-command callouts for each device." ?

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+The parameters are the same as for a "start" command. The pre-command event is
+non-disruptive in this case as to allow mdevctl to attempt each device.
```
cohuck, line 463: "For autostart devices, the return code of the pre-command callouts is ignored, and mdevctl will continue processing the list of devices even if an error has been returned for one of the devices."

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+All errors reported by the pre/post events are redirected to systemd.
```
cohuck, line 464: "are logged to the system log." ?

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+
+.PP
+\fBAuto-start Notifier: \fR
+<\fICONFIG\fR> | SCRIPT \fR
+<\fB-e=\fR\fI"notify"\fR \fB-a=\fR\fI"start"\fR
+\fB-s=\fR\fI"none"\fR|\fI"success"\fR|\fI"failure"\fR
+\fB-u=\fR\fIUUID\fR \fB-p=\fR\fIPARENT\fR>
+.RS 4
+A notifier will be invoked during the auto-start events using the "start"
+action. Note that if a notification script is used to convey information to
```
cohuck, line 521: Will that notification be invoked either before or after the primary command execution, with the same semantics as above?

mdevctl.8
```diff
@@ -382,12 +382,169 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
...
+another program or daemon during the auto-start procedure, it is not
+guaranteed that the program will be started prior to mdevctl's invocation.
```
cohuck, line 523: How would the script be invoked prior to starting mdevctl?

Makefile
cohuck: Ok, I now realize that the notifier has slightly different semantics (not tied to a specific type), so this looks good to me.

mdevctl.8
collinwalling: I meant that if the notifier script signals another program, e.g. libvirt, it's not guaranteed that the libvirt daemon would have been started prior to the udev rule firing off for mdevctl auto-start devices.

How about:

"Note that if a notification script is designed to signal another program or daemon during the auto-start procedure, it is not a guaranteed that the program/daemon will be active prior to script invocation."

mdevctl.8
collinwalling: Exactly. I can borrow the description for the generic notifier and modify it a bit.

How about:

"A notify call-out is invoked for each device during the auto-start procedure, either after a pre-command call-out failure,
or after a post-command call-out."

cohuck: Sounds good.

mdevctl.8
cohuck: Isn't it a general issue that a program that is being signalled is not neccessarily running? I would expect any program like libvirt to support some kind of scanning to figure out what actually exists when it is started.

jonner:
> That's a fair response to my poorly-worded acknowledgement.
> 
> What I meant to state was that our intention with moving forward with this pull request was to ensure that callouts will become a core feature of mdevctl and would thus be considered as a requirement for a "feature complete" version of mdevctl-experimental. I think we should discuss if that means this code gets included in the bash version and used as a basis for callouts in the rust version, or something else entirely different. This is a highly-valuable feature for vfio_ap-passthrough mdevs.

I'm sorry that the timing of my rewrite complicates this feature. Whether or not it makes sense to implement it first in the bash version and then port to the rust version or just implement directly in the new code base I'll leave as a decision for others. But I'm fine with treating this feature as a requirement for transitioning to the new code base.

> I hope these remarks don't come across as a "we did the work, now someone else can do it" stance -- definitely not my intention. 

For what it's worth, I didn't take it that way at all

> At minimum I think it's a good idea to get involved with reviewing the rust version since it has gained a lot of traction and while it's still in the early stages (though I'd have to learn the language myself). I'll confer with Matt and Joy on how we should move forward.

I'd love to have some more eyes on the rust version, and would welcome certainly feedback. I realize that the choice of language makes it a little bit more of a challenge. I am willing to help implement this feature in the rust version and/or port the bash code to rust after the design is finalized.

collinwalling:
> I'm sorry that the timing of my rewrite complicates this feature. Whether or not it makes sense to implement it first in the bash version and then port to the rust version or just implement directly in the new code base I'll leave as a decision for others. But I'm fine with treating this feature as a requirement for transitioning to the new code base.

All is well :) And thanks for acknowledging the desire for this feature.

> I'd love to have some more eyes on the rust version, and would welcome certainly feedback. I realize that the choice of language makes it a little bit more of a challenge. I am willing to help implement this feature in the rust version and/or port the bash code to rust after the design is finalized.

That sounds like a great plan and your help would be appreciated. Plus, this gives us an excuse to learn something new :) We'll need some time to catch up on the language and then we'll figure out how we can contribute. In the meantime (and if everyone is willing to) I'd like to keep working on this design until we have something solid.

mdevctl
```diff
@@ -314,6 +314,163 @@ unique_uuid() {
...
+    fi
+
+    set -o pipefail
+
+    if [ -z "$script" ]; then
+        for s in $callout_script/*; do
+            echo "$conf" | $s -t "$type" -e "$event" -a "$action" -s "$state" -u "$uuid" -p "$parent" 2>&1 > /dev/null |
+            sed "s/^/$(basename $(realpath "$s")): /" >&2
+            rc=$?
+            if [ $rc -ne 95 ]; then
```
awilliam, line 446: 95?!  I see this is also documented in the man page, but why 95?

mdevctl
```diff
@@ -314,6 +314,163 @@ unique_uuid() {
...
+                script=$s
```
awilliam, line 447: Weren't we discussing a scheme where multiple callouts could be triggered, not just the first that returned Ok/Fail?

mdevctl
```diff
@@ -1000,6 +1157,14 @@ case "$cmd" in
         ;;
 esac
 
+load_device_parameters "$uuid" "$parent" "$type" "$cmd"
+
+callout "pre" "$cmd" "$uuid" "$parent" "$type"
+if [ $? -ne 0 ]; then
+    callout_notify "$cmd" "$uuid" "$parent" "$type"
+    exit 0
```
awilliam, line 1286: Why do we exit success on callout failure?

mdevctl
```diff
@@ -314,6 +314,163 @@ unique_uuid() {
...
+    # even after the device removed from the system
+    if [ -z "$conf" ]; then
+        conf=$(dump_config)
+    fi
+
+    set -o pipefail
+
+    if [ -z "$script" ]; then
+        for s in $callout_script/*; do
+            echo "$conf" | $s -t "$type" -e "$event" -a "$action" -s "$state" -u "$uuid" -p "$parent" 2>&1 > /dev/null |
```
awilliam, line 443: I'd love a comment describing this redirection in the code so I don't need to refer to you previous review commend to understand it :)

mdevctl
collinwalling: This was my idea. It's the value for EOPNOTSUPP regarding operating system error codes. I know this error isn't coming from an OS, but I thought it might make sense to borrow from an already-defined list of error codes.

mdevctl
collinwalling: Are you referring to Conny's proposal: https://github.com/mdevctl/mdevctl/pull/32#discussion_r614762287?

From what I understand, the latter is correct. Ideally we want only one script to run per device type. Otherwise I'm not sure if it makes sense for a device type to have multiple callout scripts.

I believe undefine is the case where we might have multiple callouts during a single run of mdevctl due to device type ambiguity (e.g. `undefine -u 123` and there are multiple devices defined with uuid 123). In this case, mdevctl will iterate for each device performing a pre -> command -> post -> notify structure and clear the `script` variable at the end of each iteration since it is possible that the next device could be a different type.

cohuck: I would ideally have either 0 or 1 scripts invoked per device type. If we have a specialized script for one type and a generalized script for a group of types that covers that type as well, I'd assume the specialized script invoked the generic one (or routines shared by both.)

mdevctl
```diff
@@ -314,6 +314,163 @@ unique_uuid() {
...
+load_device_parameters() {
+    uuid="$1"
+    parent="$2"
+    type="$3"
+    cmd="$4"
+
+    set +o errexit
+
+    if [ -z "$uuid" ]; then
+        return
```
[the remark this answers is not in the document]
zhuoyingcai: This case is to skip any action for list and type commands.

mdevctl
```diff
@@ -314,6 +314,163 @@ unique_uuid() {
...
+            # We arrive here when populating $parent for start or modify.
+            # There must be only one config file exists for $uuid.
+            # Otherwise error will occur if no config or multiple configs for $uuid.
+            file_path=$(echo $(find "$persist_base" -name "$uuid" -type f))
+            parent=$(basename $(realpath "$file_path" | sed -s "s/\/$uuid//"))
+        fi
+
+        # Can't continue this function without populating $parent
+        if [ -z "$parent" ]; then
+            return 1
```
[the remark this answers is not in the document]
zhuoyingcai: Are you referring to `set -o errexit` before this function? 

If this function is executing, the return value should be checked; errexit is turned on at line 340 in this function.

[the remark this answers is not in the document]
awilliam: The return value is not checked and line 340 sets `set +o errext` to turn off exit on error, so this return 1 is quietly ignored.

mdevctl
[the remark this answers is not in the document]
awilliam: We have $cmd, wouldn't it make sense to make this less obfuscated by testing it rather than $uuid?

mdevctl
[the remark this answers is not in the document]
zhuoyingcai: What I intend to do here is return 1 here if $parent is not populated, and let callout to report any errors or exit the program. Do you have any suggestions for this? Thanks!

mdevctl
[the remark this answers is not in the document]
zhuoyingcai: :+1: I agree. Testing $cmd makes more sense.

jonner:
@collinwalling, @zhuoyingcai: I played around a little bit with porting this to my rust branch to make sure that it was possible. It's pretty quick and dirty. I looked at your bash implementation above and did a quick translation to rust and then tweaked things a little bit. I didn't really do much testing, and I'm sure that it could be refactored to be better rust code. But if you want to take a look, here's what I have at the moment: jonner/mdevctl@74af766

collinwalling:
@jonner Thanks Jonathon. I've left my comments on your WIP commit. I appreciate you taking the time to look into it.

collinwalling:
#40 is now available. Closing this PR since mdevctl is now bashless :)

PR #37, opened 2021-06-09
Fix define from jsonfile

fiuczy:
Using a jsonfile like this
$ cat mdev_nodedev.json
{"mdev_type": "vfio-ccw_io", "start": "manual", "attrs":[]}

and calling mdevctl to define a definition as follows causes an error.
$ mdevctl define -p 0.0.0033 --jsonfile mdev_nodedev.json
/usr/sbin/mdevctl: line 183: /etc/mdevctl.d/0.0.0033/532d626a-40bb-4cc1-a8f5-3291b42039ed: No such file or directory

This is due a missing subdirectory creation in the base directory
/etc/mdevctl.d/ before writing the config in the define jsonfile case.

### DIFF
--- a/mdevctl
+++ b/mdevctl
@@ -609,6 +609,7 @@ case "$cmd" in
                 exit 1
             fi
 
+            mkdir -p "$persist_base/$parent"
             write_config "$persist_base/$parent/$uuid"
             if [ $? -ne 0 ]; then
                 exit 1


cohuck:
I'm not sure whether we want to do another release of the bash implementation, even for such a straightforward fix...

Does this work for the Rust rewrite?

fiuczy:
The rust rewrite has this covered in write_config (at least as far as I understand rust).
Since this problem is hit immediately when using libvirts virsh nodedev-define command I guess that for backports into distros it would be much simpler to justify this fix in the bash version than a complete replacement.

jonner:
Yeah, the rust version doesn't have this bug. The question is whether we want to do another upstream release with bash or simply let downstream distributions backport the fix to the bash version.

awilliam:
The fix is trivial enough that I'm inclined to merge it and turn the crank for a final spin of the bash version, then we should get on with merging the rust version.

fiuczy:
> 
> 
> The fix is trivial enough that I'm inclined to merge it and turn the crank for a final spin of the bash version, then we should get on with merging the rust version.

Thanks, the likelihood to hit this bug using vfio-ccw is due to the parent-mediated device relationship of 1-to-1 much higher than for other mediated device types which usually have a 1-to-many relationship (e.g. vfio-ap).

PR #40, opened 2021-06-28
[v3] Callouts (rust impl)

collinwalling:
Version "3" of the call-out implementation. Previous version is #35 

A few notable changes from the previous version (aside from all the rust):

- scripts are now stored in /etc/mdevctl.d/scripts.d/callouts and scripts.d/notifiers
- "script not supported for device type" error code changed from 95 (loosely related to OS error codes) to 2

More documentation can be read via the man page changes, but I'll post it all here so it's easier to read without all the formatting junk:

```
EXTERNAL SCRIPTS FOR DEVICE EVENTS
   SYNOPSIS
       <CONFIG> | SCRIPT <-t=type -e=event -a=action -s=status -u=UUID -p=parent>

   DESCRIPTION
       mdevctl  supports invoking external scripts to handle additional device-specific configuration
       and event reporting. These scripts are invoked by mdevctl at various points during command
       execution depending on the "event" type and an "action". The scripts are also provided a
       "status" flag regarding the mdevctl's command success/failure (or "none" if the command was
       not executed), and the device's "TYPE", "UUID", "PARENT". The device's JSON configuration
       is provided via standard input.

       These scripts are invoked before and after an mdevctl's "primary command execution" (e.g. writing
       the device configuration file during define, or activating a device during start).

       All errors are redirected to stderr (except for auto-start call-out errors,
       which are reported to the system log).

       Essentially, the procedure in mdevctl looks like this:

              1. command-line parsing & setup

              2. invoke pre-command call-out

              3. primary command execution*

              4. invoke post-command call-out*

              5. invoke notifier

              * step is skipped if 2 fails.

   CALL-OUT EVENTS
       Call-out  event scripts are invoked with the following parameters below. For "pre", "post",
       and "get" call-outs, the "TYPE" parameter must be checked by the script to ensure the 
       device type is supported. If the script returns error code 2, denoting that the device type 
       is not supported, then mdevctl will skip the call-out event.

       Pre-Command: <CONFIG> | SCRIPT <-t=TYPE -e="pre" -a=ACTION -s="none" -u=UUID -p=PARENT>

           A pre-command call-out is invoked once prior to primary command execution.
           Action is a string denoting the mdevctl command. Status will always be "none".

           If the call-out script returns with a error code 2, mdevctl will skip the call-out event. 
           Any other non-zero return code will prevent mdevctl from performing the primary 
           command execution, and mdevctl will abort early.

           A notification event will follow only if an error occurred.

           This event is not supported for the list, types, or version commands.

       Post-Command: <CONFIG> | SCRIPT <-t=TYPE -e="post" -a=ACTION -s="success"|"failure" -u=UUID -p=PARENT>

           A post-command call-out is invoked once after to primary command execution.
           Action is a string denoting the mdevctl command. Status will be "success" if
           mdevctl was able to finish primary command execution  successfully,
           or "failure" otherwise.

           A non-zero return code is ignored.

           A notification event will always follow a post-command call-out.

           This event is not supported for the list, types, or version commands.

       Get-attributes: SCRIPT <-t=TYPE -e="get" -a="attributes" -s="none" -u=UUID -p=PARENT>

           A get call-out is invoked during a define and list command to acquire device attributes
           of an active device. Action is the string "attributes". Status will be "none". Note thatm 
           unlike other call-outs, get-attributes does  not expect a device config on stdin.

           For define, a non-zero return code will disrupt the define command entirely.

           For list, any return code will be ignored.

           A script must return a JSON formatted array of device attributes. Example output:

           [
               {
                   "attribute0": "VALUE"
               },
               {
                   "attribute1": "VALUE"
               }
           ]

   NOTIFICATION EVENT
       Notification event scripts may be used to signal the state of the mediated device or the
       status of an mdevctl command to other programs / loggers. A notification event is invoked
       with the following parameters:

       Notifier: <CONFIG> | SCRIPT <-e="notify" -a=ACTION -s="none"|"success"|"failure" -u=UUID -p=PARENT>

           A notify call-out is invoked once either after a pre-command call-out failure, or after a 
           post-command call-out. Action is a string denoting the mdevctl command. If following a
           pre-command call-out, then  status  will  be "none". If following a post-command call-out, 
           then status will be "success" if mdevctl was able to finish primary command execution 
           successfully, or "failure" otherwise.

           All installed notification scripts will be invoked during a notificaiton event.

           A non-zero return code is ignored.

           This event is not supported for the list, types, or version commands.

   AUTO-START CALL-OUTS
       For each device set to start automatically during system boot, mdevctl will invoke the pre-
       and post-command callouts. Action is the string "start".

       Return  code  and notification event behavior is the same as the pre/post-command call-outs, 
       except any error reported by a script will disrupt the auto-start for that particular device and 
       the message will be reported to the system log.

       Note that if a notification script is used to convey information to another program or daemon
       during the auto-start procedure, it is not guaranteed that the program will already be active 
       prior to mdevctl's  invocation  (e.g. auto-start may occur before the libvirt daemon is activated).

   SCRIPT RETURN VALUES
              0  if OK,

              1  if an error occurred,

              2  if the script does not support the device type

FILES
       /etc/mdevctl.d/*

       Configuration files are in one subdirectory per parent device and named by UUID.

       /etc/mdevctl.d/scripts.d/callouts/*

       Scripts for pre/post/get call-out events.

       /etc/mdevctl.d/scripts.d/notifiers/*

       Scripts for notification call-out events.
```

### DIFF
--- a/mdevctl.8
+++ b/mdevctl.8
@@ -390,12 +390,6 @@ e2e73122-cc39-40ee-89eb-b0a47d334cae matrix vfio_ap-passthrough manual
     @{2}: {"assign_domain":"0xff"}
 .EE
 
-.SH FILES
-\fI/etc/mdevctl.d/*\fR
-
-Configuration files are in one subdirectory per parent device and named
-by UUID.
-
 .SH "CONFIGURATION FILE FORMAT"
 
 Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
@@ -414,6 +408,226 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
   ]
 }
 .EE
+
+.SH INVOKING EXTERNAL SCRIPTS FOR DEVICE EVENTS
+
+mdevctl supports invoking external scripts to handle additional device
+type-specific configurations and to broadcast notifications regarding changes
+or updates to a device. These scripts are invoked before, after, and/or during
+mdevctl's "primary command execution" (e.g. writing the device configuration
+file for define, or activating a device for start).
+
+Essentially, the procedure in mdevctl looks like this:
+
+.RS
+.IP "1. command-line parsing & setup"
+.IP "2. invoke pre-command call-out"
+.IP "3. primary command execution*"
+.IP "4. invoke post-command call-out*"
+.IP "5. invoke notifier"
+.IP "* skipped if step 2 fails."
+.RE
+
+.SS EVENT SCRIPTS
+
+A call-out or notification event invokes a script along with a set of
+parameters detailing the type of call-out, mdevctl's command execution
+progress, and the mediated device. The parameters are as follows:
+
+<CONFIG> | SCRIPT <\fB-t=\fR\fItype\fR \fB-e=\fR\fIevent\fR
+\fB-a=\fR\fIaction\fR \fB-s=\fR\fIstate\fR \fB-u=\fR\fIUUID\fR
+\fB-p=\fR\fIparent\fR>
+
+.PP
+\fBCONFIG\fR
+.RS 4
+The device's JSON configuration, provided via standard input.
+.RE
+
+.PP
+\fB-t=\fR\fItype\fR
+.RS 4
+The device type.
+.RE
+
+.PP
+\fB-e=\fR\fIevent\fR
+.RS 4
+Event type of call-out that is invoked. For call-out scripts, this may be
+\fBpre\fR, \fBpost\fR, or \fBget\fR. For notification scripts, this will
+always be \fBnotify\fR.
+.RE
+
+.PP
+\fB-a=\fR\fIaction\fR
+.RS 4
+An action synonymous with an mdevctl command (e.g. define, start).
+.RE
+
+.PP
+\fB-s=\fR\fIstate\fR
+.RS 4
+A trinary state of the mdevctl command execution. The possibilities are
+\fBnone\fR if the mdevctl command has yet to execute, \fBsuccess\fR
+if the mdevctl command completed successfully, or \fBfailure\fR if there
+was a problem executing the mdevctl command.
+.RE
+
+.PP
+\fB-u=\fR\fIUUID\fR
+.RS 4
+UUID of the mediated device.
+.RE
+
+.PP
+\fB-p=\fR\fIparent\fR>
+.RS 4
+Parent of the mediated the device.
+.RE
+
+.SS CALL-OUT EVENT SCRIPTS
+
+A call-out event script is invoked during a \fBpre\fR, \fBpost\fR, or \fBget\fR
+event. mdevctl will attempt each script stored in the mdevctl callouts
+directory until either a script that satisfies the device type is found or all
+scripts have been attempted. A device script must check the "TYPE" parameter to
+ensure the specified device type is supported, otherwise error code 2 should be
+returned. If no script is found for the specified device type, then mdevctl
+will carry on as normal.
+
+These scripts are stored in \fI/etc/mdevctl.d/scripts.d/callouts\fR. The same
+script is invoked for \fBpre\fR, \fBpost\fR, and \fBget\fR call-out events for
+the device type.
+
+.PP
+\fBPre-Command\fR
+.RS 4
+
+A pre-command call-out event is invoked once prior to primary command execution.
+Event type is \fBpre\fR. Status will always be \fBnone\fR.
+
+Any non-zero return code (exempting 2) will prevent mdevctl from performing
+the primary command execution and mdevctl will abort early.
+
+A notification event will follow only if an error code (exempting 2) is
+observed.
+
+This event is not supported for the \fBlist\fR, \fBtypes\fR, or \fBversion\fR
+commands.
+.RE
+
+.PP
+\fBPost-Command\fR
+.RS 4
+
+A post-command call-out event is invoked once after primary command execution.
+Event type is \fBpost\fR. Status will be \fBsuccess\fR if mdevctl was able to
+finish primary command execution successfully, or \fBfailure\fR otherwise.
+
+The same script used for the pre event is used for the post event.
+
+Any return code is non-disruptive.
+
+A notification event will always follow a post-command call-out.
+
+This event is not supported for the \fBlist\fR, \fBtypes\fR, or \fBversion\fR
+commands.
+.RE
+
+.PP
+\fBGet-attributes\fR
+.RS 4
+
+A get event is invoked during a \fBdefine\fR and \fBlist\fR command to
+acquire device attributes from an active device. Event type is \fBget\fR. Action
+is \fBattributes\fR. Status is \fBnone\fR. Note that, unlike other call-outs
+events, \fBget-attributes does not expect a device config on stdin, and an
+array of JSON formatted device attributes is returned via stdout\fR.
+
+The same script used for the pre event is used for the get event. If the script
+is not designed to support a get event, then the return code is 0.
+
+For \fBdefine\fR, a non-zero return code (exempting 2) will disrupt the define
+command entirely.
+
+For \fBlist\fR, any return code is non-disruptive.
+
+A script must return a JSON formatted array of device attributes on standard
+output. Example:
+
+.EX
+[
+    {
+        \fI"attribute0"\fR: \fI"VALUE"\fR
+    },
+    {
+        \fI"attribute1"\fR: \fI"VALUE"\fR
+    }
+]
+.EE
+.RE
+
+.SS AUTO-START CALL-OUTS
+
+For each device set to start automatically during system boot, mdevctl will
+invoke the pre and post events. Action is the string \fBstart\fR.
+
+Return code and notification event behavior is the same as documented for
+the pre and post events. Errors reported by a script will disrupt the
+auto-start for that particular device and the message will be reported to the
+system log before attempting to the next auto-start device.
+
+Note that if a notification script is used to convey information to another
+program or daemon during the auto-start procedure, it is not guaranteed that
+the program will already be active prior to mdevctl's invocation (e.g. the
+auto-start event may occur before the libvirt daemon is activated).
+.RE
+
+.SS NOTIFICATION EVENT SCRIPTS
+
+Notification event scripts may be used to signal the state of the mediated
+device or the status of an mdevctl command to other programs or loggers. Unlike
+call-out scripts, notifier scripts are device-type agnostic.
+
+.PP
+\fBNotify\fR
+.RS 4
+
+A notification event is invoked once either following a pre-command call-out
+failure or after a post-command call-out. Event is \fBnotify\fR. If following a
+pre event, then status will be \fBnone\fR. If following a post event, then
+status will mirror the value passed to the post-command call-out.
+
+These scripts are stored in \fI/etc/mdevctl.d/scripts.d/notifiers\fR. \fBAll
+notification scripts will be invoked during a notification event\fR.
+
+A non-zero return code is ignored.
+
+This event is not supported for the \fBlist\fR, \fBtypes\fR, or \fBversion\fR
+commands.
+.RE
+
+.SS SCRIPT RETURN VALUES
+
+.RS
+.IP "0  if OK,
+.IP "1  if an error occurred,
+.IP "2  if the script does not support the device type
+.RE
+
+.SH FILES
+\fI/etc/mdevctl.d/*\fR
+
+Configuration files are in one subdirectory per parent device and named
+by UUID.
+
+\fI/etc/mdevctl.d/scripts.d/callouts/*\fR
+
+Scripts for pre/post/get call-out events.
+
+\fI/etc/mdevctl.d/scripts.d/notifiers/*\fR
+
+Scripts for notification call-out events.
 
 .SH "SEE ALSO"
 \fBudev\fR(7)
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -13,6 +13,7 @@ pub enum Event {
     Pre,
     Post,
     Notify,
+    Get,
 }
 
 impl Display for Event {
@@ -27,6 +28,9 @@ impl Display for Event {
             Event::Notify => {
                 write!(f, "notify")
             }
+            Event::Get => {
+                write!(f, "get")
+            }
         }
     }
 }
@@ -38,6 +42,7 @@ pub enum Action {
     Define,
     Undefine,
     Modify,
+    Attributes,
 }
 
 impl Display for Action {
@@ -48,6 +53,7 @@ impl Display for Action {
             Action::Define => write!(f, "define"),
             Action::Undefine => write!(f, "undefine"),
             Action::Modify => write!(f, "modify"),
+            Action::Attributes => write!(f, "attributes"),
         }
     }
 }
@@ -107,6 +113,52 @@ impl Callout {
         res
     }
 
+    pub fn get_attributes(dev: &mut MDev) -> Result<serde_json::Value> {
+        let event = Event::Get;
+        let action = Action::Attributes;
+        let c = Callout::new();
+        let dir = dev.env.callout_script_base();
+
+        if !dir.is_dir() {
+            return Ok(serde_json::Value::Null);
+        }
+
+        match c.invoke_first_matching_script(dev, dir, event, action) {
+            Some((path, output)) => {
+                if output.status.success() {
+                    debug!("Get attributes successfully from callout script");
+                    let mut st = String::from_utf8_lossy(&output.stdout).to_string();
+
+                    if st.is_empty() {
+                        return Ok(serde_json::Value::Null);
+                    }
+
+                    if &st == "[{}]" {
+                        debug!(
+                            "Attribute field for {} is empty",
+                            dev.uuid.to_hyphenated().to_string()
+                        );
+                        st = "[]".to_string();
+                    }
+
+                    serde_json::from_str(&st)
+                        .with_context(|| anyhow!("Unable to parse attributes from JSON"))
+                } else {
+                    c.print_err(&output, &path);
+
+                    return Err(anyhow!("failed to get attributes from {:?}", path));
+                }
+            }
+            None => {
+                debug!(
+                    "Device type {} unmatched by callout script",
+                    dev.mdev_type.as_ref().unwrap()
+                );
+                Ok(serde_json::Value::Null)
+            }
+        }
+    }
+
     fn invoke_script<P: AsRef<Path>>(
         &self,
         dev: &mut MDev,
@@ -139,12 +191,13 @@ impl Callout {
             .stderr(Stdio::piped());
 
         let mut child = cmd.spawn()?;
-
-        let conf = dev.to_json(false)?.to_string();
-        if let Some(mut child_stdin) = child.stdin.take() {
-            child_stdin
-                .write_all(conf.as_bytes())
-                .with_context(|| "Failed to write to stdin of command")?;
+        if event != Event::Get {
+            let conf = dev.to_json(false)?.to_string();
+            if let Some(mut child_stdin) = child.stdin.take() {
+                child_stdin
+                    .write_all(conf.as_bytes())
+                    .with_context(|| "Failed to write to stdin of command")?;
+            }
         }
 
         child.wait_with_output().map_err(anyhow::Error::from)
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -36,6 +36,7 @@ impl Display for Event {
 }
 
 #[derive(Clone, Copy)]
+#[allow(dead_code)]
 pub enum Action {
     Start,
     Stop,
@@ -43,6 +44,7 @@ pub enum Action {
     Undefine,
     Modify,
     Attributes,
+    Test, // used for tests only
 }
 
 impl Display for Action {
@@ -54,6 +56,7 @@ impl Display for Action {
             Action::Undefine => write!(f, "undefine"),
             Action::Modify => write!(f, "modify"),
             Action::Attributes => write!(f, "attributes"),
+            Action::Test => write!(f, "test"),
         }
     }
 }
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -12,6 +12,7 @@ use crate::mdev::*;
 pub enum Event {
     Pre,
     Post,
+    Notify,
 }
 
 impl Display for Event {
@@ -22,6 +23,9 @@ impl Display for Event {
             }
             Event::Post => {
                 write!(f, "post")
+            }
+            Event::Notify => {
+                write!(f, "notify")
             }
         }
     }
@@ -84,7 +88,7 @@ impl Callout {
     {
         let mut c = Callout::new();
 
-        c.callout(dev, Event::Pre, action).and_then(|_| {
+        let res = c.callout(dev, Event::Pre, action).and_then(|_| {
             let tmp_res = func(dev);
             c.state = match tmp_res {
                 Ok(_) => State::Success,
@@ -97,7 +101,10 @@ impl Callout {
             }
 
             tmp_res
-        })
+        });
+
+        let _ = c.notify(dev, action);
+        res
     }
 
     fn invoke_script<P: AsRef<Path>>(
@@ -225,5 +232,32 @@ impl Callout {
                 n
             )),
         }
+    }
+
+    fn notify(&mut self, dev: &mut MDev, action: Action) -> Result<()> {
+        let event = Event::Notify;
+        let dir = dev.env.callout_notification_base();
+
+        if !dir.is_dir() {
+            return Ok(());
+        }
+
+        for s in dir.read_dir()? {
+            let path = s?.path();
+
+            match self.invoke_script(dev, &path, event, action) {
+                Ok(output) => {
+                    if !output.status.success() {
+                        debug!("Error occurred when executing notify script {:?}", path);
+                    }
+                }
+                _ => {
+                    debug!("Failed to execute callout script {:?}", path);
+                    continue;
+                }
+            }
+        }
+
+        Ok(())
     }
 }
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -1,0 +1,229 @@
+use anyhow::{anyhow, Context, Result};
+use log::{debug, warn};
+use std::ffi::OsStr;
+use std::fmt::{self, Display, Formatter};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Output, Stdio};
+
+use crate::mdev::*;
+
+#[derive(Clone, Copy, PartialEq)]
+pub enum Event {
+    Pre,
+    Post,
+}
+
+impl Display for Event {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            Event::Pre => {
+                write!(f, "pre")
+            }
+            Event::Post => {
+                write!(f, "post")
+            }
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+pub enum Action {
+    Start,
+    Stop,
+    Define,
+    Undefine,
+    Modify,
+}
+
+impl Display for Action {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            Action::Start => write!(f, "start"),
+            Action::Stop => write!(f, "stop"),
+            Action::Define => write!(f, "define"),
+            Action::Undefine => write!(f, "undefine"),
+            Action::Modify => write!(f, "modify"),
+        }
+    }
+}
+
+#[derive(Clone, Copy)]
+pub enum State {
+    None,
+    Success,
+    Failure,
+}
+
+impl Display for State {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            State::None => write!(f, "none"),
+            State::Success => write!(f, "success"),
+            State::Failure => write!(f, "failure"),
+        }
+    }
+}
+
+pub struct Callout {
+    state: State,
+    script: Option<PathBuf>,
+}
+
+impl Callout {
+    fn new() -> Callout {
+        Callout {
+            state: State::None,
+            script: None,
+        }
+    }
+
+    pub fn invoke<F>(dev: &mut MDev, action: Action, func: F) -> Result<()>
+    where
+        F: Fn(&mut MDev) -> Result<()>,
+    {
+        let mut c = Callout::new();
+
+        c.callout(dev, Event::Pre, action).and_then(|_| {
+            let tmp_res = func(dev);
+            c.state = match tmp_res {
+                Ok(_) => State::Success,
+                Err(_) => State::Failure,
+            };
+
+            let post_res = c.callout(dev, Event::Post, action);
+            if post_res.is_err() {
+                debug!("Error occurred when executing post callout script");
+            }
+
+            tmp_res
+        })
+    }
+
+    fn invoke_script<P: AsRef<Path>>(
+        &self,
+        dev: &mut MDev,
+        script: P,
+        event: Event,
+        action: Action,
+    ) -> Result<Output> {
+        debug!(
+            "{} callout: executing {:?}",
+            event.to_string(),
+            script.as_ref().as_os_str()
+        );
+
+        let mut cmd = Command::new(script.as_ref().as_os_str());
+
+        cmd.arg("-t")
+            .arg(dev.mdev_type()?)
+            .arg("-e")
+            .arg(event.to_string())
+            .arg("-a")
+            .arg(action.to_string())
+            .arg("-s")
+            .arg(self.state.to_string())
+            .arg("-u")
+            .arg(dev.uuid.to_string())
+            .arg("-p")
+            .arg(dev.parent()?)
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped());
+
+        let mut child = cmd.spawn()?;
+
+        let conf = dev.to_json(false)?.to_string();
+        if let Some(mut child_stdin) = child.stdin.take() {
+            child_stdin
+                .write_all(conf.as_bytes())
+                .with_context(|| "Failed to write to stdin of command")?;
+        }
+
+        child.wait_with_output().map_err(anyhow::Error::from)
+    }
+
+    fn print_err<P: AsRef<Path>>(&self, output: &Output, script: P) {
+        let sname = script
+            .as_ref()
+            .file_name()
+            .unwrap_or_else(|| OsStr::new("unknown script name"))
+            .to_string_lossy();
+
+        let st = String::from_utf8_lossy(&output.stderr);
+        if !st.is_empty() {
+            eprint!("{}: {}", &sname, st);
+        }
+    }
+
+    fn invoke_first_matching_script<P: AsRef<Path>>(
+        &self,
+        dev: &mut MDev,
+        dir: P,
+        event: Event,
+        action: Action,
+    ) -> Option<(PathBuf, Output)> {
+        if dir.as_ref().read_dir().ok()?.count() == 0 {
+            return None;
+        }
+
+        for s in dir.as_ref().read_dir().ok()? {
+            let path = s.ok()?.path();
+
+            match self.invoke_script(dev, &path, event, action).ok() {
+                Some(res) => {
+                    if res.status.code().is_none() {
+                        warn!("Script was terminated by a signal");
+                        continue;
+                    } else if res.status.code() != Some(2) {
+                        return Some((path, res));
+                    } else {
+                        debug!(
+                            "Device type {} unmatched by callout script",
+                            dev.mdev_type().ok()?
+                        );
+                    }
+                }
+                _ => {
+                    debug!("failed to execute callout script {:?}", path);
+                    continue;
+                }
+            }
+        }
+        None
+    }
+
+    fn callout(&mut self, dev: &mut MDev, event: Event, action: Action) -> Result<()> {
+        let rc = match self.script {
+            Some(ref s) => self
+                .invoke_script(dev, s, event, action)
+                .ok()
+                .and_then(|output| {
+                    self.print_err(&output, s);
+                    output.status.code()
+                }),
+            _ => {
+                let dir = dev.env.callout_script_base();
+
+                if !dir.is_dir() {
+                    return Ok(());
+                }
+                self.invoke_first_matching_script(dev, dir, event, action)
+                    .and_then(|(path, output)| {
+                        self.print_err(&output, &path);
+                        self.script = Some(path);
+                        output.status.code()
+                    })
+            }
+        };
+
+        match rc {
+            Some(0) | None => Ok(()),
+            Some(n) => Err(anyhow!(
+                "callout script {:?} failed with return code {}",
+                self.script.as_ref().unwrap(),
+                n
+            )),
+        }
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,11 +16,13 @@ use std::vec::Vec;
 use structopt::StructOpt;
 use uuid::Uuid;
 
+use crate::callouts::*;
 use crate::cli::{LsmdevOptions, MdevctlCommands};
 use crate::environment::{DefaultEnvironment, Environment};
 use crate::logger::logger;
 use crate::mdev::*;
 
+mod callouts;
 mod cli;
 mod environment;
 mod logger;
@@ -133,8 +135,9 @@ fn define_command(
 ) -> Result<()> {
     debug!("Defining mdev {:?}", uuid);
 
-    let dev = define_command_helper(env, uuid, auto, parent, mdev_type, jsonfile)?;
-    dev.define().map(|_| {
+    let mut dev = define_command_helper(env, uuid, auto, parent, mdev_type, jsonfile)?;
+
+    Callout::invoke(&mut dev, Action::Define, |dev| dev.define()).map(|_| {
         if uuid.is_none() {
             println!("{}", dev.uuid.to_hyphenated());
         }
@@ -149,8 +152,8 @@ fn undefine_command(env: &dyn Environment, uuid: Uuid, parent: Option<String>) -
         return Err(anyhow!("No devices match the specified uuid"));
     }
     for (_, mut children) in devs {
-        for child in children.iter_mut() {
-            child.undefine()?;
+        for mut child in children.iter_mut() {
+            let _ = Callout::invoke(&mut child, Action::Undefine, |dev| dev.undefine());
         }
     }
     Ok(())
@@ -193,7 +196,7 @@ fn modify_command(
         }
     }
 
-    dev.write_config()
+    Callout::invoke(&mut dev, Action::Modify, |dev| dev.write_config())
 }
 
 /// convert 'start' command arguments into a MDev struct
@@ -287,7 +290,8 @@ fn start_command(
     jsonfile: Option<PathBuf>,
 ) -> Result<()> {
     let mut dev = start_command_helper(env, uuid, parent, mdev_type, jsonfile)?;
-    dev.start().map(|_| {
+
+    Callout::invoke(&mut dev, Action::Start, |dev| dev.start()).map(|_| {
         if uuid.is_none() {
             println!("{}", dev.uuid.to_hyphenated());
         }
@@ -299,7 +303,8 @@ fn stop_command(env: &dyn Environment, uuid: Uuid) -> Result<()> {
     debug!("Stopping '{}'", uuid);
     let mut dev = MDev::new(env, uuid);
     dev.load_from_sysfs()?;
-    dev.stop()
+
+    Callout::invoke(&mut dev, Action::Stop, |dev| dev.stop())
 }
 
 /// convenience function to lookup a defined device by uuid and parent
--- a/src/main.rs
+++ b/src/main.rs
@@ -668,7 +668,7 @@ fn start_parent_mdevs_command(env: &dyn Environment, parent: String) -> Result<(
         for child in children {
             if child.autostart {
                 debug!("Autostarting {:?}", child.uuid);
-                if let Err(e) = child.start() {
+                if let Err(e) = Callout::invoke(child, Action::Start, |child| child.start()) {
                     for x in e.chain() {
                         warn!("{}", x);
                     }
--- a/src/main.rs
+++ b/src/main.rs
@@ -137,6 +137,16 @@ fn define_command(
 
     let mut dev = define_command_helper(env, uuid, auto, parent, mdev_type, jsonfile)?;
 
+    /*
+        Call Callout::get_attributes() when defining an active device without a config file.
+        This function allows callout script to acquire device-specific attributes from sysfs,
+        and populate the attrs field correspondingly before the device is defined in the system.
+        The device config file will contain the same attributes that were used to start this device。
+    */
+    if dev.active {
+        let attrs = Callout::get_attributes(&mut dev)?;
+        dev.add_attributes(&attrs)?;
+    }
     Callout::invoke(&mut dev, Action::Define, |dev| dev.define()).map(|_| {
         if uuid.is_none() {
             println!("{}", dev.uuid.to_hyphenated());
@@ -478,6 +488,12 @@ fn list_command_helper(
                     }
 
                     let _ = dev.load_definition();
+
+                    if !dev.is_defined() {
+                        if let Ok(attrs) = Callout::get_attributes(&mut dev) {
+                            let _ = dev.add_attributes(&attrs);
+                        }
+                    }
 
                     let devparent = dev.parent()?;
                     if !devices.contains_key(devparent) {
--- a/src/mdev.rs
+++ b/src/mdev.rs
@@ -25,7 +25,7 @@ pub struct MDev<'a> {
     pub parent: Option<String>,
     pub mdev_type: Option<String>,
     pub attrs: Vec<(String, String)>,
-    env: &'a dyn Environment,
+    pub env: &'a dyn Environment,
 }
 
 impl<'a> MDev<'a> {
--- a/src/mdev.rs
+++ b/src/mdev.rs
@@ -207,9 +207,9 @@ impl<'a> MDev<'a> {
 
         let mut output = self.uuid.to_hyphenated().to_string();
         output.push(' ');
-        output.push_str(&self.parent()?);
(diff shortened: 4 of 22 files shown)


logger.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/bash
+# A basic utility script used for debugging notification call-out events
+# in mdevctl. Output can be observed via system logs.
+#
+# Place this script in /etc/mdevctl.d/scripts.d/notifiers/ and run
```
collinwalling, line 5: Of course there will always be _something_ I miss. The script path mentioned on the man page and the comment here are correct. The commit message is incorrect.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde_json = {version = "1.0", features = ["preserve_order"]}
 structopt = "0.3.13"
 uuid = {version = "0.8.2", features = ["v4"]}
 tempdir = "0.3"
+libsystemd = "0.3.0"
```
cohuck, line 22: I'm wondering whether we could use an alternative logging method that does not depend on libsystemd.

mdevctl.8
```diff
@@ -407,6 +401,185 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
...
+\fB-a=\fR\fIaction\fR \fB-s=\fR\fIstatus\fR \fB-u=\fR\fIUUID\fR
+\fB-p=\fR\fIparent\fR>
+
+.SS DESCRIPTION
+
+mdevctl supports invoking external scripts to handle additional device-specific
+configuration and event reporting. These scripts are invoked by mdevctl at
+various points during command execution depending on the "event" type and an
+"action". The scripts are also provided a "status" flag regarding the mdevctl's
+command success/failure (or "none" if the command was not executed), and the
```
cohuck, line 417: Maybe
```
The scripts are also provided a "status" flag signifying the success respectively failure of the
mdevctl command
```

mdevctl.8
```diff
@@ -407,6 +401,185 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
...
+.IP "5. invoke notifier"
+.IP "* step is skipped if 2 fails."
+.RE
+
+.SS CALL-OUT EVENTS
+Call-out event scripts are invoked with the following parameters below. For
+"pre", "post", and "get" call-outs, the "TYPE" parameter must be checked by
+the script to ensure the device type is supported. If the script returns
+error code 2, denoting that the device type is not supported, then mdevctl will
+skip the call-out event.
```
cohuck, line 444: The last sentence is a bit confusing: will mdevctl continue with the next script?

mdevctl.8
```diff
@@ -407,6 +401,185 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
...
+<\fICONFIG\fR> | SCRIPT \fR
+<\fB-t=\fR\fITYPE\fR \fB-e=\fR\fI"pre"\fR \fB-a=\fR\fIACTION\fR
+\fB-s=\fR\fI"none"\fR \fB-u=\fR\fIUUID\fR \fB-p=\fR\fIPARENT\fR>
+.RS 4
+
+A pre-command call-out is invoked once prior to primary command execution.
+Action is a string denoting the mdevctl command. Status will always be "none".
+
+If the call-out script returns with a error code 2, mdevctl will skip the
+call-out event. Any other non-zero return code will prevent mdevctl from
```
cohuck, line 457: Same as above.

mdevctl.8
```diff
@@ -407,6 +401,185 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
...
+.PP
+\fBGet-attributes: \fR
+SCRIPT \fR
+<\fB-t=\fR\fITYPE\fR \fB-e=\fR\fI"get"\fR \fB-a=\fR\fI"attributes"\fR
+\fB-s=\fR\fI"none"\fR \fB-u=\fR\fIUUID\fR \fB-p=\fR\fIPARENT\fR>
+.RS 4
+
+A get call-out is invoked during a \fBdefine\fR and \fBlist\fR command to
+acquire device attributes of an active device. Action is the string
+"attributes". Status will be "none". Note thatm unlike other call-outs,
```
cohuck, line 496: s/thatm/that/

mdevctl.8
collinwalling: Yes, mdevctl will continue with the next script. What if I change the last sentence to:

```
The script must return error code 2 if it does not support the device type. mdevctl will
attempt the next script in the callouts directory until either a different return code is 
observed or all scripts have been exhausted.
```

mdevctl.8
collinwalling: "signifying the success respectively failure of the mdevctl command" doesn't read too well to me (maybe I'm not awake yet and can't process it right now).

I assume you'd prefer more elaboration on the trinary values? How about:

```
The scripts are also provided a "status" flag signifying the success or failure of the 
respective mdevctl command. For pre-command call-outs a status of "none" will be
reported, as the command has yet to executed.
```

jonner (commented):
Just wanted to note that `cargo test` fails for me on this branch. I haven't investigated the failures yet.

zhuoyingcai:
> Just wanted to note that `cargo test` fails for me on this branch. I haven't investigated the failures yet.

This branch passed `cargo test` after populating the callout script directories in the device environment when the TestEnvironment is initialized. I've fixed this on another patch.

collinwalling:
Force pushed changes to address above:

- documentation suggestions
- Joy's fix for `cargo test` failures

mdevctl.8
```diff
@@ -407,6 +401,186 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
...
+
+.PP
+\fBPost-Command: \fR
+<\fICONFIG\fR> | SCRIPT \fR
+<\fB-t=\fR\fITYPE\fR \fB-e=\fR\fI"post"\fR \fB-a=\fR\fIACTION\fR
+\fB-s=\fR\fI"success"\fR|\fI"failure"\fR \fB-u=\fR\fIUUID\fR
+\fB-p=\fR\fIPARENT\fR>
+.RS 4
+
+A post-command call-out is invoked once after to primary command execution.
```
jonner, line 475: to -> the?

mdevctl.8
```diff
@@ -407,6 +401,186 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
...
+.RS 4
+
+A notify call-out is invoked once either after a pre-command call-out failure,
+or after a post-command call-out. Action is a string denoting the mdevctl
+command. If following a pre-command call-out, then status will be "none". If
+following a post-command call-out, then status will be "success" if mdevctl
+was able to finish primary command execution successfully, or "failure"
+otherwise.
+
+All installed notification scripts will be invoked during a notificaiton event.
```
jonner, line 540: spelling of notification

jonner (commented):
This looks like a reasonable first step. One big thing to pay attention to is the use of `unwrap()`. I noted several cases, but not all of them. Be careful to only use this if you can absolutely guarantee that it will not panic.

jonner:
Also, please run `cargo fmt` before every commit to ensure consistent coding style.

src/mdev.rs
```diff
@@ -629,4 +630,55 @@ impl<'a> Callout<'a> {
...
+                } else {
+                    return Err(anyhow!("failed to get attributes from {:?}", self.script));
+                }
+            }
+        }
+
+        let re = Regex::new(r"\w+").unwrap();
+        for cap in re.captures_iter(&st) {
+            v.push(cap[0].to_string());
+        }
```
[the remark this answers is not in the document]
zhuoyingcai: The output of the callout script for get is an array of attributes. For example [{"assign_adapter": "2"},{"assign_domain": "1"}]. I'm not sure if serde_json::from_str() is useful here.

src/mdev.rs
```diff
@@ -480,3 +481,110 @@ impl MDevType {
...
+        let mut rc = Some(0);
+
+        let dir = self.mdev.env.callout_script_base();
+        if dir.read_dir()?.count() == 0 {
+            return Ok(());
+        }
+
+        if self.script.to_str().unwrap().is_empty() {
+            for s in dir.read_dir()? {
+                let path = &s?.path();
```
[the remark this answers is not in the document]
zhuoyingcai: I tried to remove the reference here. But I'm getting a move error.
```
# cargo build
   Compiling mdevctl v1.0.0 (/root/mdevctl-rust)
error[E0382]: use of moved value: `path`
   --> src/mdev.rs:608:40
    |
604 |                 let path = s?.path();
    |                     ---- move occurs because `path` has type `PathBuf`, which does not implement the `Copy` trait
605 |                 let res = self.invoke_script(path, event, action, true, false, self.mdev.autostart);
    |                                              ---- value moved here
...
608 |                     self.script = Some(path);
    |                                        ^^^^ value used here after move

error: aborting due to previous error

For more information about this error, try `rustc --explain E0382`.
error: could not compile `mdevctl`

To learn more, run the command again with --verbose.
```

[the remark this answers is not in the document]
jonner: Since the `path` variable is no longer a reference, you'll need to take a reference when passing it to the invoke_script() function.

`self.invoke_script(&path, ...)`

src/mdev.rs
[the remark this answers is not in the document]
jonner: Yes, it is an array of attributes in JSON format, so it makes sense to parse it with a JSON parser. This is essentially what I had in mind (NOTE: NOT TESTED)

````diff
diff --git a/src/main.rs b/src/main.rs
index d39dced..7603ee4 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -695,11 +695,9 @@ where F: FnMut(&mut MDev) -> Result<()>, {
 fn invoke_callout_get(dev: &mut MDev) -> Result<()> {
     let mut c = Callout::new(dev.clone());
 
-    let vec = c.callout_get("get", "attributes")?;
+    let attrs = c.callout_get("get", "attributes")?;
 
-    for i in (0..vec.len()).step_by(2) {
-        dev.add_attribute(vec[i].clone(), vec[i+1].clone(), None)?;
-    }
+    dev.add_attributes(&attrs);
 
     Ok(())
 }
diff --git a/src/mdev.rs b/src/mdev.rs
index 9e66c86..5f03af9 100644
--- a/src/mdev.rs
+++ b/src/mdev.rs
@@ -11,7 +11,6 @@ use std::vec::Vec;
 use uuid::Uuid;
 use std::process::{Command, Stdio, Output, ExitStatus};
 use libsystemd::logging::{self, Priority};
-use regex::Regex;
 
 #[derive(Clone, Copy)]
 pub enum FormatType {
@@ -126,6 +125,20 @@ impl<'a> MDev<'a> {
         Ok(())
     }
 
+    pub fn add_attributes(&mut self, attrs: &serde_json::Value) {
+        if let Some(attrarray) = attrs.as_array() {
+            if !attrarray.is_empty() {
+                for attr in attrarray {
+                    let attrobj = attr.as_object().unwrap();
+                    for (key, val) in attrobj.iter() {
+                        let valstr = val.as_str().unwrap();
+                        self.attrs.push((key.to_string(), valstr.to_string()));
+                    }
+                }
+            }
+        }
+    }
+
     pub fn load_from_json(&mut self, parent: String, json: &serde_json::Value) -> Result<()> {
         debug!(
             "Loading device '{:?}' from json (parent: {})",
@@ -156,17 +169,7 @@ impl<'a> MDev<'a> {
         let startval = json["start"].as_str();
         self.autostart = matches!(startval, Some("auto"));
 
-        if let Some(attrarray) = json["attrs"].as_array() {
-            if !attrarray.is_empty() {
-                for attr in json["attrs"].as_array().unwrap() {
-                    let attrobj = attr.as_object().unwrap();
-                    for (key, val) in attrobj.iter() {
-                        let valstr = val.as_str().unwrap();
-                        self.attrs.push((key.to_string(), valstr.to_string()));
-                    }
-                }
-            }
-        };
+        self.add_attributes(&json["attrs"]);
         debug!("loaded device {:?}", self);
 
         Ok(())
@@ -631,14 +634,13 @@ impl<'a> Callout<'a> {
         Ok(())
     }
 
-    pub fn callout_get(&mut self, event: &str, action: &str) -> Result<Vec<String>> {
-        let mut v: Vec<String> = Vec::new();
+    pub fn callout_get(&mut self, event: &str, action: &str) -> Result<serde_json::Value> {
         let mut st = String::new();
 
         let dir = self.mdev.env.callout_script_base();
 
         if dir.read_dir()?.count() == 0 {
-            return Ok(v);
+            return Ok(Default::default());
         }
 
         for s in dir.read_dir()? {
@@ -674,11 +676,6 @@ impl<'a> Callout<'a> {
             }
         }
 
-        let re = Regex::new(r"\w+").unwrap();
-        for cap in re.captures_iter(&st) {
-            v.push(cap[0].to_string());
-        }
-
-        Ok(v)
+        serde_json::from_str(&st).with_context(|| anyhow!("Unable to parse attributes from JSON"))
     }
 }
````

**Comment by jonner** [comment]
Also, have you put any thought into how we might test this functionality?

**Comment by collinwalling** [comment]
> Also, have you put any thought into how we might test this functionality?

We have some internal tooling for a vfio_ap-passthrough script that is still in the works and tethers some mdevctl testing as well, but obviously that isn't sufficient for this project.

I'm currently working on a callout script test framework. Basically it will have a script template that will handle the option parsing (to make sure mdevctl is passing the correct data) and some handlers that will pass back expected results. It'll also return some bogus data back as well to make sure mdevctl catches it.

Additionally, the script template could be used as a starting point for anyone who would like to implement their own script in rust.

File: src/mdev.rs [hunk 11]
```diff
@@ -570,15 +581,15 @@ impl<'a> Callout<'a> {
         if self.script.to_str().unwrap().is_empty() {
             for s in dir.read_dir()? {
                 let path = &s?.path();
-                let res = self.invoke_script(path, event, action, true, false);
+                let res = self.invoke_script(path, event, action, true, false, self.mdev.autostart);
```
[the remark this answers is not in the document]
zhuoyingcai: At this point, the error from the script is printing to both stderr and system log for an autostart device if it's being operated manually. The error will only show in system log when the autostart devices are starting at boot time.

Is there a way to determine whether an autostart device is being operated manually?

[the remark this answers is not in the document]
jonner: At the moment there's not really any way to determine that. But I could imagine e.g. adding an additional argument to `invoke_with_callback(..., log_errors: bool)`, and storing that value within the `Callout` struct. This parameter would only be set to `true` when called by `start_parent_mdevs_command()`. 

I'm not sure that this is the most elegant solution, you may be able to think of something better.

zhuoyingcai:
Change log v1→v2:
1) Fix compiler error in Add callout function to mdevctl commit.
2) Make the `sname` field in the Callout struct as a local variable in `Callout::print_output()`.
3) Change the type of `script` field in the Callout struct from `PathBuf` to `Option<PathBuf>`.
4) Remove reference from `DirEntry::path()`.
5) Change the type of `conf` field in the Callout struct from `String` to `Option<String>`.
6) Move the Callout struct out of mdev.rs module and make it a separate module in callouts.rs.
7) Add argument, `log_errors: bool`, to `invoke_with_callout()`, and this value is stored in the Callout struct. This parameter is set to `true` when `start_parent_mdevs_command()` is invoked, and print warnings to the system log.
8) Simplify command block for ‘notify’ event.
9) Create `enums` for events and convert them to a string when passing to the script.
10) Avoid use `unwarp()`.
11) Change return value of `Callout::callout_get()` to `Result<serde_json::Value>`. Factor out the part of `MDev::load_from_json()` that handles the attributes and re-use that code to insert attributes from `Callout::callout_get()` to a mdev device.
12) Fix bug where an empty attribute field is retrieved by the callout script.

zhuoyingcai:
Collin is working on the callouts testing. He will be back Tuesday 7/20 and will have more to share then.

src/callouts.rs
```diff
@@ -0,0 +1,174 @@
...
+            .as_ref()
+            .ok_or_else(|| anyhow!("Failed to get callout script path"))
+    }
+
+    fn match_event(&self, event: EventType) -> &str {
+        match event {
+            EventType::Pre => "pre",
+            EventType::Post => "post",
+        }
+    }
```
jonner: Consider implementing [`ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html) for `EventType` instead of adding a member function to Callout. The `self` parameter is not even used here. If you implement `ToString`,  you can just do something like `event.to_string()`

jonner (commented):
I've just looked through the first commit so far. It's getting closer.

src/callouts.rs
```diff
@@ -190,4 +196,24 @@ impl<'a> Callout<'a> {
             )),
         }
     }
+
+    pub fn callout_notify(
```
jonner, line 202: Since the function is already within the `Callout` struct, I don't think that the `callout_` prefix is necessary. `Callout::notify()` would seem sufficiently descriptive.

src/callouts.rs
```diff
@@ -190,4 +196,24 @@ impl<'a> Callout<'a> {
...
+    ) -> Result<()> {
+        let dir = env.callout_notification_base();
+
+        if dir.read_dir()?.count() == 0 {
+            return Ok(());
+        }
+
+        for s in dir.read_dir()? {
+            let path = s?.path();
+            let _ = self.invoke_script(&path, event, action, true, true);
```
jonner, line 216: Out of curiosity, what's the reason for the difference in the `stderr`, `stdout` arguments between the normal callouts and the notify callout?

src/main.rs
```diff
@@ -713,6 +713,7 @@ where
         Err(e) => Err(e),
     };
 
+    let _ = c.callout_notify(env, EventType::Notify, action);
```
jonner, line 733: I wonder if it's worth at least printing a warning or debug message to the terminal if there was an error executing the notify script? Now that I think about it, perhaps the same is true of line 710 above.

jonner (commented):
2nd patch. Also, commit message has a typo: "retults" -> "results"

Cargo.toml
jonner: So, I just did a little bit of looking at it appears that you should be able to use something like the [syslog](https://docs.rs/syslog/5.0.0/syslog/) crate and it would avoid the dependency on systemd. The question is: how does it integrate with the existing env_logger logging framework that we're using? I haven't looked that far.

src/callouts.rs
```diff
@@ -18,6 +19,7 @@ pub struct Callout<'a> {
     state: &'a str,
     conf: Option<String>,
     script: Option<PathBuf>,
+    log_errors: bool,
```
jonner, line 24: the `log_errors` name is a bit inaccurate since we're logging both stdout and stderr, not just stderr. Also, even when `log_errors` is false, we still log the errors, we just log them to the terminal, right? So maybe something like `use_syslog` is a better description?

src/main.rs
```diff
@@ -717,6 +734,16 @@ where
     res
 }
 
+fn invoke_callout_get(env: &dyn Environment, dev: &mut MDev) -> Result<()> {
+    let mut c = Callout::new(dev.clone());
+
+    let attrs = c.callout_get(env, EventType::Get, "attributes")?;
```
jonner, line 740: So, we construct a `Callout` object `c`, and then call the `c.callout_get()` function, but then the `c` object simply goes out of scope. So this is just a slightly more verbose way of calling a standalone function with an extra `dev` argument: e.g. `callout_get(dev, env, EventType::Get, "attributes")`. Maybe that's fine, but it suggests to me that perhaps there's a better way to organize this code...

src/callouts.rs
```diff
@@ -216,4 +218,76 @@ impl<'a> Callout<'a> {
 
         Ok(())
     }
+
+    pub fn callout_get(
+        &mut self,
+        env: &dyn Environment,
+        event: EventType,
+        action: &str,
+    ) -> Result<serde_json::Value> {
```
jonner, line 227: Looking over this again, it seems that there's a fair bit of logic here is similar to stuff in `callout()`. Is there any way we can refactor to share stuff? Maybe that can be done later as a follow-up.

src/main.rs
```diff
@@ -137,11 +137,24 @@ fn define_command(
 
     let mut dev = define_command_helper(env, uuid, auto, parent, mdev_type, jsonfile)?;
 
-    invoke_with_callout(env, &mut dev, false, "define", |dev| dev.define()).map(|_| {
-        if uuid.is_none() {
-            println!("{}", dev.uuid.to_hyphenated());
+    if dev.active {
+        match invoke_callout_get(env, &mut dev) {
```
jonner, line 141: It seems like this deserves some code comments about why we should call invoke the callout_get when the device is active.

src/mdev.rs
```diff
@@ -123,6 +123,31 @@ impl<'a> MDev<'a> {
         Ok(())
     }
 
+    pub fn add_attributes(&mut self, attrs: &serde_json::Value) -> Result<()> {
+        if let Some(attrarray) = attrs.as_array() {
```
jonner, line 127: Now that we've exposed this function to be used by other callers, perhaps we should return an error if the json value is not an array.

jonner (commented):
4th patch

src/callouts.rs
jonner: and I realize that I was the one that suggested this name ;)

src/callouts.rs
collinwalling: For pre/post callouts, `stdout` is suppressed to avoid any extraneous noise on the command line.

For `get attributes`, mdevctl will expect a JSON array of device attributes on `stdout`. (note: this is implemented via `invoke_callout_get`)

For `notify`, I think we can focus a discussion on that. It is not 100% fleshed out what notifier scripts should do, other than potentially echo results to another program. @jonner , I imagine your libvirt work could potentially make use of a notifier (i.e. when an `mdevctl define` completes, a notifier can echo the results to a `virsh nodev-create` or something like that)? I'm not sure if notify would need to echo anything back to mdevctl via `stdout` here. What do you think? @cohuck?

If we can't come up with a reasonable use-case for mdevctl to output `stdout` from a notifier script, then I say we should set the `stdout` flag to false for notifiers.

jonner: I can't really think of any situation where `stdout` from a notification script would be useful. In fact, I'm not sure whether `stderr` from a notification script is something we'd want to log. For the pre- and post- callbacks, it makes sense to log error output from scripts because that will help explain why the mdevctl operation might have failed. But notification scripts shouldn't affect the mdevctl operation at all, so printing printing an error message from a random notification script (which the user might not even be aware of) will probably just confuse the user. They might interpret it as the mdevctl command itself failing when it actually succeeded and only the random notification script failed. If many scripts are installed, It could end up spewing a output to the terminal. That doesn't seem like a very good user experience.

Perhaps there should be a way to enable printing of output from notification scripts to allow script developers to debug their notification scripts (e.g. with an env variable or something), but I'm starting to think that these should both be false by default.

src/callouts.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+        if stdout {
+            let st = String::from_utf8_lossy(&output.stdout);
+            if !st.is_empty() {
+                let s = format!("{}: {}", &sname, st);
+
+                if !self.log_errors {
+                    print!("{}", &s);
+                } else {
+                    let _ = logging::journal_print(Priority::Warning, &s);
```
jonner, line 150: I think `Priority::Warning` is probably wrong for logging `stdout` output. Probably `Notice` or `Info` or `Debug` is better.

src/callouts.rs
cohuck: I agree. If the notification script does not influence the command, we should not add confusion by logging whatever it says; I'd consider notification to be one-way.
A debug switch might be helpful, if it's easy to implement.

src/callouts.rs
cohuck: I'd say Notice or Info.

src/callouts.rs
collinwalling: That makes sense to me.

I'm wondering if we should go one step further with refining this interface. Why don't we keep the parameters the same for every event type? The only difference between the callouts and notifiers are the absence of the `type` parameter for notify. We could simplify the code (and the documentation) if we just include the type parameter for notify. We could also treat `stdout` for callouts as a way of communicating information _back_ to mdevctl (e.g. for get). The code could be simplified to have a single callout function where the only difference is the event type, action, etc., toggling the capture of `stdout`, and explicitly ignoring `stderr` for notify. Those special cases can be handled in a wrapper function.

jonner: @collinwalling yep, that sounds reasonable to me.

zhuoyingcai:
Change log v2 → v3
1) Handle error when getting output from `Path::file_name()`.
2) Use `OsStr::to_string_lossy()` to avoid error causing by non-Unicode file name.
3) Return Output from `Callout::invoke_script()` and store it in Callout struct. So the output can be later used in other functions in Callout struct.
4) Pass `MDev` object as a reference to `Callout::callout()`.
5) Avoid cloning the child when calling `invoke_with_callout()` on undefine.
6) Implement `ToString` for `EventType`.
7) Print debug message to the terminal if there was an error executing the notify script and post callout script.
8) Change variable name `log_errors` to `use_syslog`.
9) Add comments about why we should call `invoke_callout_get()` when the device is active.
10) Return an error if the JSON value is not an array in `Mdev::add_attributes()`.
11) Change priority from `Warning` to `Notice` for `logging::journal_print()` when printing stdout from callou scripts.
11) Refactor and simplify callout interface so that different events is invoked through a single `Callout::callout()` interface.

collinwalling:
I have some progress on the testing effort callouts. It took me a bit longer than I'd like to admit to figure out how to organize it all, but I have a working presentation available here: https://github.com/zhuoyingcai/mdevctl/tree/callouts-testing

The code includes a "script template" library with a struct that parses command line opts using structopts and a trait that defines a shell for all pre/post_command variations. The library also includes a `run_script` function that reads in the command line opts, reads in stdin (if required), checks if the command line type and parent matches that of what the script expects, then drills down to the appropriate pre/post_command function. Using this library is easy:

```
use scripttemplate::*;

struct Script;

impl ScriptFunctions for Script {
    fn notify(&self, opts: ScriptOpts) -> i32 {
        println!("logger_script: {:?}", opts);
        return 0
    }
}

fn main() {
    run_script_and_exit("", "", &Script);
}
```
The empty strings are for `mdev_type` and `parent`, but in this example, any type and parent are accepted. A script only needs to implement the functions they need. By default, the functions simply return 0.

As for the testing efforts, I have two basic scripts that implement this trait. One that has all functions return 0 (no actual changes), and one that has all functions return 1. Then, `tests.rs` will test the `invoke_with_callout` function, calling out to the built scripts and anticipating a pass or fail.

The main issue I had when designing all of this was figuring out "how to relocate the cargo-built callout scripts to the tests data directory". For this iteration, I have these scripts in the `examples` directory, which will get built when running `cargo test`. The `tests.rs` file will look in `target/debug/examples/<filename>` for the appropriate script to run. This probably isn't the best approach, but I'm not sure what the best "post build" process would be to move those scripts into the tests data directory.

@jonner , thanks for your patience and I'd appreciate any-and-all feedback you have to offer. This is my first attempt at rust traits and making a library, so any advice on best practices / styling is welcome.

jonner:
@collinwalling I'll try to take a look at your testing stuff soon. I'm not sure if I'll have time this week or not. It looks interesting.

mdevctl.8
```diff
@@ -415,6 +409,187 @@ Configuration files are in JSON. Attributes in \fB"attrs"\fR are optional.
...
+Action is a string denoting the mdevctl command. Status will be "success" if
+mdevctl was able to finish primary command execution successfully, or "failure"
+otherwise.
+
+A non-zero return code is ignored.
+
+A notification event will always follow a post-command call-out.
+
+This event is not supported for the \fBlist\fR, \fBtypes\fR, or \fBversion\fR
+commands.
```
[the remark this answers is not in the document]
collinwalling: Good point. I think I'll rework this to have a "Device Script" section that puts emphasis on "the same script for the device type" with a subsection for pre/post combined (detailing the differences, of course), get, and auto-start. Notifier will have its own section as it does already.

jonner:
@collinwalling I'm not sure whether there's a better place to discuss this, but I'll just put it here for now. I think your callout script library looks like a nice start. I'm also not particularly experienced with libraries in rust, so we'll both learn as we go I guess ;)

I think there's probably some things I would change about the API but we can discuss specifics later. By the way, what is the 'Test' event? I don't remember discussion of that.

With regard to testing specifically, it also looks like a pretty good start. It's basic at the moment, but I see you have a few things listed in a TODO list in the commit log. Right now you're currently testing what happens when a callout script returns a success or a failure for all possible events, but obviously there's quite a bit more that could be tested that's not listed in the TODO.

For instance, off the top of my head, we could:

1. test that the appropriate callout script gets chosen when you run it for a particular mdev_type
2. test that the appropriate values are passed to the callout script
3. test that mdevctl can handle a misbehaving callout script (e.g. segfault, deadlock)
4. test that the post callout doesn't get called if the pre callout returns a failure
5. test that notify callout happens regardless of success or failure and that all notification scripts are executed regardless of callout script return value
6. test that we can handle odd filenames, permissions problems
7. test that we can gracefully handle bad JSON output from the get attributes callout

Obviously, we don't need to expend all of our energy testing every conceivable possibility right away, but adding tests for some of these scenarios will help ensure that the infrastructure works so that we don't introduce regressions that break the callout functionality in the future.

collinwalling:
> @collinwalling I'm not sure whether there's a better place to discuss this, but I'll just put it here for now. I think your callout script library looks like a nice start. I'm also not particularly experienced with libraries in rust, so we'll both learn as we go I guess ;)

Sounds good to me!
> 
> I think there's probably some things I would change about the API but we can discuss specifics later. By the way, what is the 'Test' event? I don't remember discussion of that.
> 

I tossed that in last minute with the idea that it may make sense to test the `callout` function instead of the `invoke_with_callout` function. If testing the former, we can change the callout.rs code to accept a Test event type, and the test scripts would only have to implement a single "test" function. If testing the latter, the test scripts would need to cover both the pre/post operations for an arbitrary command, which can be a bit cumbersome (the tests would also have to implement a function that gets called inbetween the callouts).

In short, I think it would be easiest/cleanest to test the `callouts` function versus the `invoke_with_callout` function.

> With regard to testing specifically, it also looks like a pretty good start. It's basic at the moment, but I see you have a few things listed in a TODO list in the commit log. Right now you're currently testing what happens when a callout script returns a success or a failure for all possible events, but obviously there's quite a bit more that could be tested that's not listed in the TODO.
> 
> For instance, off the top of my head, we could:
> 
>     1. test that the appropriate callout script gets chosen when you run it for a particular mdev_type
> 
>     2. test that the appropriate values are passed to the callout script
> 
>     3. test that mdevctl can handle a misbehaving callout script (e.g. segfault, deadlock)
> 
>     4. test that the post callout doesn't get called if the pre callout returns a failure
> 
>     5. test that notify callout happens regardless of success or failure and that all notification scripts are executed regardless of callout script return value
> 
>     6. test that we can handle odd filenames, permissions problems
> 
>     7. test that we can gracefully handle bad JSON output from the get attributes callout
> 
> 
> Obviously, we don't need to expend all of our energy testing every conceivable possibility right away, but adding tests for some of these scenarios will help ensure that the infrastructure works so that we don't introduce regressions that break the callout functionality in the future.

These are all great ideas.

zhuoyingcai:
Change log v3 → v4
1) Remove caching the `attrs`, `conf`, and `output` from the `Callout` struct.
2) Remove unnecessary variable accessors.
3) Use normal error handling to avoid aborting the program due to a missing script name.
4) Return `Ok(())` when script directory doesn't exist.
5) Simplify match statement when invoking `Callout:get_attributes()`.
6) Make `env` variable public in `MDev` struct.
7) Add error-handling to make sure faulty callout/notify scripts (eg. not executable, wrong permission, corrupt file, etc) will not block the mdevctl command. 
8) Continue checking the next script if the current script has `None` return code or did not execute successfully.
9) Ignore `None` return values in `Callout::callout()`. 
10) Use `enum` for mdevctl action and callout state.
11) Move `invoke()` and `get_attributes()` to `Callout` struct.
12) Remove needless borrow for Clippy in mdev.rs.

I'm still looking into the syslog change.

src/callouts.rs
```diff
@@ -0,0 +1,354 @@
...
+        let dir = dev.env.callout_script_base();
+
+        if !dir.is_dir() {
+            return Ok(());
+        }
+
+        let rc;
+
+        if let Some(ref s) = self.script {
+            let output = self.invoke_script(dev, s, event, action)?;
```
jonner, line 304: Here, I think we want to avoid using the `?` operator and returning an error from `callout()` if `invoke_script()` fails. We only want to return an error (and block primary command execution) if the script successfully executes and returns a valid response code. This prevents buggy scripts from disrupting mdevctl's functionality.

src/callouts.rs
```diff
@@ -0,0 +1,354 @@
...
+                }
+                None => rc = None,
+            }
+        }
+
+        match rc {
+            Some(0) | None => Ok(()),
+            _ => Err(anyhow!(
+                "aborting command due to results from callout script {:?}",
+                self.script.as_ref().unwrap()
```
jonner, line 322: Inside of this function, you don't actually know whether the command will be aborted or not. The caller (`Callout::invoke()`) is the one who decides whether to abort the command or not. So perhaps we should use a slightly different message. For example: 

```suggestion
            Some(n) => Err(anyhow!(
                "callout script {:?} failed with return code {}",
                self.script.as_ref().unwrap(), n
```

(adding the return code to the error message might help script writers debug failures, but it might not be the best user experience...)

src/callouts.rs
```diff
@@ -0,0 +1,354 @@
...
+
+                let post_res = c.callout(dev, Event::Post, action);
+                if post_res.is_err() {
+                    debug!("Error occurred when executing post callout script");
+                }
+
+                tmp_res
+            }
+            Err(e) => Err(e),
+        };
```
jonner, line 120: Any time you have a branch that just maps from the input to an identical output (`Err(e) => Err(e)`), there's probably a better way to do things. In this case, I'd suggest something like this:

```suggestion
        let res = c.callout(dev, Event::Pre, action).and_then(|_| {
            let tmp_res = func(dev);
            c.state = match tmp_res {
                Ok(_) => State::Success,
                Err(_) => State::Failure,
            };

            let post_res = c.callout(dev, Event::Post, action);
            if post_res.is_err() {
                debug!("Error occurred when executing post callout script");
            }

            tmp_res
        });

```

src/callouts.rs
```diff
@@ -0,0 +1,354 @@
...
+        if !dir.is_dir() {
+            return Ok(());
+        }
+
+        for s in dir.read_dir()? {
+            let path = s?.path();
+
+            match self.invoke_script(dev, &path, event, action) {
+                Ok(output) => {
+                    self.print_output(&output, &path, false, false);
```
jonner, line 340: if both `stderr` and `stdout` are `false`, then `print_output()` does nothing. Maybe remove this line?.

jonner (changes requested):
Thanks for your patience. It's getting very close now.

src/callouts.rs
```diff
@@ -0,0 +1,354 @@
...
+            }
+            Err(e) => return Err(anyhow!("{}", e)),
+        }
+    }
+
+    fn print_output<P: AsRef<Path>>(&self, output: &Output, script: P, stderr: bool, stdout: bool) {
+        let sname = script
+            .as_ref()
+            .file_name()
+            .unwrap_or_else(|| OsStr::new("unknow script name"))
```
jonner, line 228:
```suggestion
            .unwrap_or_else(|| OsStr::new("unknown script name"))
```

zhuoyingcai:
Change log v4 → v5
1) Handle errors more concisely. 
2) Handle errors for post callout to avoid buggy scripts.
3) Change the error message when an error occurs in callout script.
4) Remove `print_output()` from `Callout::notify()`.
5) Fix typo for "unknown script name".
6) Use `syslog` crate to avoid dependency on `systemd` (send outputs of callout script to journalctl successfully, but I haven't figure out how to integrate it with the `env_logger`. Please let me know if you have any suggestions. Thank you!)

jonner (commented):
The few remaining comments are pretty minor. It seems about ready to merge, though it would be nice to have a few tests to give a little more confidence that we're handling the corner cases properly.

src/main.rs
```diff
@@ -676,11 +705,37 @@ fn start_parent_mdevs_command(env: &dyn Environment, parent: String) -> Result<(
...
+                }
+                _ => return start_parent_mdevs_command(&env, parent),
+            }
+        }
+        _ => {
+            logger().init();
+            debug!("Starting up");
+            debug!("{:?}", env);
+        }
+    }
```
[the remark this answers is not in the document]
jonner: Here is my experiment: https://github.com/jonner/mdevctl/commit/0bb241f6c0997df8f7af9c810f6907257427a61c

That commit applies on top of your branch. What do you think?

cohuck:
FWIW, I currently don't have any further comments to add; so I guess when @jonner is happy, I'm happy :)

collinwalling:
> The few remaining comments are pretty minor. It seems about ready to merge, though it would be nice to have a few tests to give a little more confidence that we're handling the corner cases properly.

Would you prefer test cases included in this PR or as a follow-up (where we can focus the discussion on the tests)?

I've been working down a list of test cases using the method I mentioned previously. They work, but they certainly need some refinement.

jonner:
> Would you prefer test cases included in this PR or as a follow-up (where we can focus the discussion on the tests)?

@collinwalling In an ideal world the tests would accompany the feature, but we can be flexible. Up until the rust re-write, there weren't any tests, after all. How much testing have you done to ensure this feature meets your needs? I've not done much so far beyond code inspection, and I don't know as much about the specific use-cases you're designing this for.

collinwalling:
So far I'm only testing the `Callouts::invoke` and `get_attributes` functions. For pre/post, have callout scripts that return each expected return code, tests that the callouts pass the expected parameters to each mdevctl command, observe what happens when there is no script for the device type. For get attributes, I have tests that return expected and bad JSON.

Testing `Callouts::invoke` is a tad tricky, since we only observe a return code from the pre callout. Post and notify execute silently. I may need to modify the callout code slightly so the tests can be more inclusive.

src/main.rs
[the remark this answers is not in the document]
zhuoyingcai: I think the code looks great! It's much cleaner than before and works as expected.

jonner:
Indeed, this is a tricky feature to test thoroughly. Some things are easier to test from the mdevctl side (e.g. "Do we handle the output from a script and do the right thing with it?"), and some are easier to test from the script side (e.g. "did I get executed when I expected and with the right parameters?"). In my mind, the highest priority at first is to ensure that the new feature doesn't disrupt existing functionality or cause regressions, especially when scripts are buggy or poorly behaved. Introducing a new feature that doesn't quite work yet is not as bad as introducing a new feature that breaks existing functionality.

collinwalling:
> In my mind, the highest priority at first is to ensure that the new feature doesn't disrupt existing functionality or cause regressions, especially when scripts are buggy or poorly behaved. 

Solid point. Right now, my tests are isolated from the other tests you've implemented. I wonder if it makes sense to implement a way re-run all of your tests with a "dummy" script installed to make sure everything works? What I'm thinking is having a test that first installs a script that simply returns success in all cases to make sure all the commands run as expected then re-runs all the current tests. I'm just running this idea off the top of my head at the moment -- does that make sense at all?

src/main.rs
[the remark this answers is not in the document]
jonner: By the way, I'm not totally sure whether 'fern' is the right answer here or not. It doesn't seem to have a lot of recent activity, so I'm not sure how well-maintained it is. But it does seem to provide the functionality we need in a fairly simple way. Most of the other crates looked too simple or too complicated. 

One other minor issue is that the fern syslog integration only supports syslog v3 or v4. But the latest release of syslog is v5, and this is the only version that is packaged for fedora. That doesn't matter if you're just building from source, but it means that it will possibly be a bit more difficult to get mdevctl packaged in fedora if we switch to fern. There's been a bug open with an attached PR to add support for v5 for over a year (https://github.com/daboross/fern/issues/47), but it hasn't seen much activity.

src/callouts.rs
```diff
@@ -113,6 +119,48 @@ impl Callout {
...
+        let c = Callout::new();
+        let dir = dev.env.callout_script_base();
+
+        if !dir.is_dir() {
+            return Ok(serde_json::Value::Null);
+        }
+
+        match c.invoke_first_matching_script(dev, dir, event, action) {
+            Some((path, output)) => {
+                if output.status.success() {
```
collinwalling, line 134: Found a bug associated with this callout. If the script returns 0 with nothing on stdout, then this function will still attempt to parse the empty output and result in an error. This is unwanted behavior, as a script may simply return 0 to indicate "yes this is the correct script for the device type, no I do not support get_attributes".

We need to either implement a new error code that signifies the above, or have the callouts code handle this case. I suggest the latter as it makes it easier for script designers to not need concern themselves with implementing an empty get_attributes function that simply returns this new error code. I'm open to the former or alternatives.

jonner:
> I'm just running this idea off the top of my head at the moment -- does that make sense at all?

hmm, I was going to say that it might make sense but that I realized that most of the tests don't call the actual command function. In other words, my tests don't call `define_command()`, they call `define_command_helper()` so that I can examine the returned device object, etc. Because of this, most of the tests will not actually invoke any callouts. So rerunning the existing test suite after installing a callout won't actually do much. 

For the purposes of testing this feature, we don't actually need to test any specific mdevctl command, we just need to test the generic `Callout::invoke()` function. It doesn't matter what we pass as the `func` argument -- it could just be an arbitrary function that sets a bit to indicate whether it was called or not. So I think it might be sufficient for now to add a new `#[test]` function that just does something like this pseudocode (not tested, but hopefully you get the idea):

```rust
fn callout_helper(callout: Option<PathBuf>) -> bool {
    if callout.is_some() {
        install_in_test_environment(callout);
    }
    let mut dev = make_device();
    let called = false;
    let _ = Callout::invoke(&mut dev, false, Action::Start, |_| {called = true});

   return called;
}

#[test]
fn test_callouts() {
    // these should all pass
    assert!(callout_helper(None)); // no callouts installed
    assert!(callout_helper(callout_that_returns_success));
    assert!(callout_helper(callout_that_returns_skip));
    assert!(callout_helper(callout_that_segfaults));
    assert!(callout_helper(callout_without_permissions));

    // this should fail
    assert!(!callout_helper(callout_that_returns_error));
}
````

File: src/callouts.rs (hunk 29 above)
Comment (collinwalling) [reply src/callouts.rs:134]: > I did not see this behavior with `define` (so maybe the list code needs adjustment).

Ah, this is because the test cases invoke the `define_command_helper` The get attributes callout is invoked in the `define_command` proper function, so this behavior was not observed during the define tests. I highly suspect the same behavior exists for this command.

Comment (jonner) [reply src/callouts.rs:134]: I agree that we should handle a 0 return code with an empty string as meaning there are no attributes. There's no need to introduce a new error code.

**Comment by collinwalling** [comment]
Yeah, that's very close to what I have currently. The `test_callouts` function is structured more like the other tests (passing in parameters for Pass/Fail, test names, a function for setting up the scripts, etc), and end with an assert:

```
fn test_callout<F>(testname: &str, 
                              expect: Expect, 
                              action: Action, 
                              uuid: Option<Uuid>, 
                              mdev_type: &str, 
                              parent: &str, 
                              setupfn: F)
where
    F: Fn(&TestEnvironment),
{
    let test = TestEnvironment::new("invoke_callout_pre_post", testname);
    setupfn(&test);

    let mut empty_mdev = MDev::new(&test, uuid.unwrap());
    empty_mdev.mdev_type = Some(mdev_type.to_string());
    empty_mdev.parent = Some(parent.to_string());

    let res = Callout::invoke(&mut empty_mdev, false, action, |_empty_mdev|  Ok(()));

    if expect == Expect::Fail {
        res.expect_err("expected callout to fail");
        return;
    }

    assert!(res.is_ok());
}
```

and a test case is invoked as such:

```
    test_callout(
        "test_callout_all_success",
        Expect::Pass,
        Action::Test,
        Uuid::parse_str(DEFAULT_UUID).ok(),
        "test_type",
        "test_parent",
        |test| {
            test.populate_callout_script("callout_test_all_pass");
        },
    );
```
I need to refine / reduce this a bit. Currently, all callout_tests just use the same UUID and Action. The type and parent parameters come into play when testing how mdevctl operates when it needs to try multiple scripts to find the right one for the device type.

> In other words, my tests don't call define_command(), they call define_command_helper() so that I can examine the returned device object, etc.

Perfect timing, as I _just_ observed this behavior myself :P. List seems to be the exception, as I found some issues there when I tried the tests.

**Comment by jonner** [comment]
Yep, that looks pretty good to me.

File: src/main.rs (hunk 28 above)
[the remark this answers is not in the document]
Comment (zhuoyingcai) [reply src/main.rs:737]: Since there might be an issue getting mdevctl packaged in Fedora if we switch to fern, do you think it's a good idea to stick with the current approach (using syslog crate only)?

@cohuck  Hi Conny, do you have any suggestions?

**Comment by collinwalling** [comment]
Unless I'm mistaken, only the `define` and `start` tests explicitly call the helper functions, and the rest of the command tests call the "actual" command functions. If this is the case, then I don't think it's too much trouble to define a couple of test functions to call the `define_command` and `start_command` for the sake of adding test coverage for the callouts code. Reminder that `list` is the exception to this case.

I'm aiming to have something presentable by tomorrow.

File: src/main.rs (hunk 28 above)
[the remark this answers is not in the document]
Comment (cohuck) [reply src/main.rs:737]: Hm, I'm not sure whether we want to depend on something that is not very active; but I find it hard to judge the quality of fern (though it seems to be a good fit for what we want in principle.)
I guess I'll chicken out and defer to @jonner for the decision :)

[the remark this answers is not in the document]
Comment (jonner) [reply src/main.rs:737]: I just had a small realization. The reason that we want `start-parent-mdevs` to log the messages to the syslog is not really because of anything special about that command itself. It's simply because it is being called by udev non-interactively so there's no way to see the output unless it is logged somewhere. If we ran this command manually, we'd still want mdevctl to print any error messages to the terminal so that the user could see them.

So in my mind, it would be ideal if mdevctl Could continue to log all output to stdout (or stderr) and have the caller (udev, in this case) capture the output and send it to the syslog . Is there any way to specify this in a udev rule? @awilliam or @cohuck: do either of you know? If we can make udev handle the syslog stuff, then we don't need to add any additional logging infrastructure within mdevctl.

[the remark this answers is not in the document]
Comment (awilliam) [reply src/main.rs:737]: @jonner, we're running mdevctl from udev with a /bin/sh -c command, so I don't see why we couldn't just pipe it to logger as part of that command.  We could even be fancy and pipe stdout and stderr w/ different priority levels, ex: sh -c '{ mdevctl start-parent-mdevs %k 2>&3 | logger -t mdevctl; } 3>&1 1>&2 | logger -t mdevctl -p 2'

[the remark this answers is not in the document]
Comment (jonner) [reply src/main.rs:737]: @awilliam I had considered something like that, but wasn't sure if there was anything more elegant built into udev. After reading some of the udev code, my impression was "probably not", but I thought I'd ask. 

I'm leaning toward simply piping the output to the syslog within the udev rule rather than complicating the mdevctl code with multiple loggers, unless someone wants to convince me otherwise.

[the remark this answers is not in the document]
Comment (collinwalling) [reply src/main.rs:737]: The udev changes would be cleaner and simpler. I'd say let's try it out.

**Comment by collinwalling** [comment]
Here is an update to the testing efforts. Things are still a bit minimal, but there is more coverage than previously: https://github.com/zhuoyingcai/mdevctl/commits/callouts-testing

Took a bit longer than expected. Enjoy the weekend!

**Review by jonner — COMMENTED** [review]
I think this is basically ready to merge with a couple minor changes. The main open question is whether define should actually proceed if `get_attributes()` fails / returns invalid JSON. I'm not sure I know the right answer.

File: src/tests.rs [hunk 30]
```diff
@@ -415,6 +454,51 @@ fn test_define() {
...
+        "define-with-callout-all-bad-json",
+        Expect::Pass,
+        Uuid::parse_str(DEFAULT_UUID).ok(),
+        Some(DEFAULT_PARENT.to_string()),
+        Some("i915-GVTg_V5_4".to_string()),
+        |test| {
+            test.populate_active_device(DEFAULT_UUID, DEFAULT_PARENT, "i915-GVTg_V5_4");
+            test.populate_callout_script("bad-json.sh");
+        },
+    );
```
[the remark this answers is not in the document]
collinwalling: I'll correct this behavior. A `define` should definitely fail if a bad json response is returned.

src/tests.rs
```diff
@@ -1332,3 +1489,209 @@ fn test_types() {
         Some("missing".to_string()),
     );
 }
+
+fn test_invoke_callout<F>(
+    testname: &str,
+    expect: Expect,
+    action: Action,
+    uuid: Option<Uuid>,
```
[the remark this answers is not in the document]
collinwalling: Are you certain? `.ok()` returns an `Option<T>`, so I think it's still needed in the arguments?

src/callouts.rs
```diff
@@ -0,0 +1,326 @@
...
+                );
+                Ok(serde_json::Value::Null)
+            }
+        }
+    }
+
+    pub fn get_attributes(dev: &mut MDev) -> serde_json::Value {
+        match Callout::get_attributes_helper(dev) {
+            Ok(j) => j,
+            Err(_e) => serde_json::Value::Null,
```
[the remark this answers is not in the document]
collinwalling: It probably would've helped if I bothered to read my own documentation too...

src/tests.rs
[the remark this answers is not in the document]
jonner: I'm just saying that if you passed a value of `None` for `uuid` in this function, it would crash because you just `.unwrap()` it. So you should make the function requirements explicit by just passing the value rather than an `Option`. You can replace the `.ok()` with `.unwrap()` in the calling function.

collinwalling:
Changelog v6 -> v7:

- reverted "get_attributes" code
  - `define` will fail on a bad json response from "get_attributes" callout
  - `list` is non-disruptive when a bad json response is received
- tests cleanups
  - list+callouts tests use their own aptly named test files
  - `Option` removed from callout helpers; callout helper invocations now use `.unwrap()`
  - updated callout test expectations based on "get_attributes" changes above
  - added missing newlines to bash scripts
- minor cleanup in `callout` function

jonner (approved):
This looks good. Thanks!

collinwalling:
Excellent work @zhuoyingcai ! Thanks for your time and review @jonner . Thanks to @cohuck @awilliam , Matt, and others for helping with the design.

PR #41, opened 2021-06-29
Fix unexpected behavior with 'start'

### LINKED ISSUE #38: starting a device with parent, type, and uuid options will ignore an existing device config (exists in both bash and rust)
When issuing `mdevctl start --parent --type --uuid` in an attempt to start a mediated device with a predefined configuration file with the same UUID, mdevctl will completely ignore the file and quietly create + start a "bare" device. This behavior is present in both the bash and rust versions.

Looking at the man page for the start command:
```
       start DEVICESPEC
           Start  an  mdev  device,  identified by its UUID and optionally its
           parent, or its parent and type and optionally its  UUID,  which  is
           generated if not given.  If specified via its parent and optionally
           its UUID, the type may be specified in a JSON  configuration  file,
           alongside additional parameters.
```
This _does_ say to read the JSON config, specify the parent and optionally a UUID. But does it really make sense to quietly ignore a JSON file if a type is also specified? It may be that at one point, if you had the UUID, type, parent and manual/auto you had everything -- but that's not true for devices with attributes.

In this case, let's say I have a vfio-ap device config that includes attributes at `/etc/mdevctl.d/matrix/2e098a15-e131-4dff-93af-2404eb7e5ddb`:
```
{
  "mdev_type": "vfio_ap-passthrough",
  "start": "auto",
  "attrs": [
    {
      "assign_domain": "0x3b"
    },
    {
      "assign_adapter": "2"
    }
  ]
}
```
If I run `mdevctl start -u 2e098a15-e131-4dff-93af-2404eb7e5ddb`, the JSON file is read in and the device is started as expected.

If I run `mdevctl start -t vfio_ap-passthrough -p matrix -u 2e098a15-e131-4dff-93af-2404eb7e5ddb`, no JSON read is attempted; it's quietly ignored in favor of a new device with no attributes. With debugging enabled (rust version), the following is observed:
```
1         [2021-06-24T19:12:20Z DEBUG mdevctl] Starting up
2         [2021-06-24T19:12:20Z DEBUG mdevctl] DefaultEnvironment { rootpath: "/" }
3         [2021-06-24T19:12:20Z DEBUG mdevctl] Starting device 'Some(2e098a15-e131-4dff-93af-2404eb7e5ddb)'
4         [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Creating mdev 2e098a15-e131-4dff-93af-2404eb7e5ddb
5         [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Loading device '2e098a15-e131-4dff-93af-2404eb7e5ddb' from sysfs
6         [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] loaded device MDev { uuid: 2e098a15-e131-4dff-93af-2404eb7e5ddb, active: false, autostart: false, parent: None, mdev_type: None, attrs: [], env: Environment { \
mdev_base: "/sys/bus/mdev/devices", persist_base: "/etc/mdevctl.d", parent_base: "/sys/class/mdev_bus" } }
7         [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Checking parent for mdev support: "/sys/class/mdev_bus/matrix/mdev_supported_types"
8         [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Checking parent for mdev type vfio_ap-passthrough: "/sys/class/mdev_bus/matrix/mdev_supported_types/vfio_ap-passthrough"
9         [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Checking available instances: "/sys/class/mdev_bus/matrix/mdev_supported_types/vfio_ap-passthrough/available_instances"
10        [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Available instances: 65536
11        [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Creating mediated device: 2e098a15-e131-4dff-93af-2404eb7e5ddb -> "/sys/class/mdev_bus/matrix/mdev_supported_types/vfio_ap-passthrough/create"
12        [2021-06-24T19:12:20Z DEBUG mdevctl::mdev] Setting attributes for mdev 2e098a15-e131-4dff-93af-2404eb7e5ddb
```
Where we can see on line 6, autostart is false (should be true), and the attributes array is empty (should show an adapter and domain). Additionally, a mediated device with this UUID has been quietly created and activated sans attributes -- confirmed by outputting the contents of the device's matrix in sysfs, which shows up empty:
```
# cat /sys/devices/vfio_ap/matrix/2e098a15-e131-4dff-93af-2404eb7e5ddb/matrix 
# 
```
The key thing there being **quietly** -- we finally figured out what was going on by enabling DEBUG on the rust version and noting that it wasn't reading anything from JSON. We can easily rectify this by providing less information on the mdevctl start -- but is there any reason mdevctl ignores a possible JSON file when type and parent are both specified? Shouldn't mdevctl rather read it optionally?  Or at least give some sort of indication that a new device was created VS something from JSON?

While this may be the desirable outcome for attributeless devices, certain devices (such as vfio_ap-passthrough) would benefit from pulling device attributes from an already-defined config file with the same UUID.

jonner:
When running `start` and providing a uuid, parent, and type that match
an existing device definition, mdevctl was creating a new device and
starting that rather than starting the defined device. This normally
doesn't matter unless the defined device contains attributes, in which
case the newly started device won't have the expected attributes set.

Also add some tests to ensure correct behavior in these scenarios.

Fixes #38

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,6 +7,7 @@
 
 use anyhow::{anyhow, ensure, Context, Result};
 use log::{debug, warn};
+use std::cmp::Ordering;
 use std::collections::BTreeMap;
 use std::fs;
 use std::io::Read;
@@ -225,22 +226,36 @@ fn start_command_helper(
             dev = Some(d);
         }
         _ => {
-            if mdev_type.is_some() && parent.is_none() {
-                return Err(anyhow!("can't provide type without parent"));
-            }
-
-            /* The device is not fully specified without TYPE, we must find a config file, with optional
-             * PARENT for disambiguation */
-            if mdev_type.is_none() {
-                if let Some(uuid) = uuid {
-                    dev = match get_defined_device(env, uuid, parent.as_ref()) {
-                        Ok(d) => Some(d),
-                        Err(e) => return Err(e),
-                    };
+            // if the user specified a uuid, check to see if they're referring to a defined device
+            if uuid.is_some() {
+                let devs = defined_devices(env, uuid.as_ref(), parent.as_ref())?;
+                let n = devs.values().flatten().count();
+                match n.cmp(&1) {
+                    Ordering::Greater => {
+                        return Err(anyhow!(
+                            "Multiple definitions found for device {}. Please specify a parent.",
+                            uuid.unwrap().to_hyphenated().to_string()
+                        ));
+                    }
+                    Ordering::Equal => {
+                        let d = devs.values().flatten().next();
+                        if let Some(d) = d {
+                            // See https://github.com/mdevctl/mdevctl/issues/38
+                            // If a user specifies the uuid (and optional parent) of a defined device
+                            if mdev_type.is_some() && mdev_type != d.mdev_type {
+                                return Err(anyhow!(
+                                    "Device {} already exists on parent {} with type {}",
+                                    d.uuid.to_hyphenated().to_string(),
+                                    d.parent().unwrap(),
+                                    d.mdev_type.as_ref().unwrap()
+                                ));
+                            } else {
+                                dev = Some(d.clone());
+                            }
+                        }
+                    }
+                    _ => (),
                 }
-            }
-            if uuid.is_none() && (parent.is_none() || mdev_type.is_none()) {
-                return Err(anyhow!("Device is insufficiently specified"));
             }
 
             if dev.is_none() {
@@ -248,6 +263,15 @@ fn start_command_helper(
                 d.parent = parent;
                 d.mdev_type = mdev_type;
                 dev = Some(d);
+            }
+
+            if let Some(ref d) = dev {
+                if d.mdev_type.is_some() && d.parent.is_none() {
+                    return Err(anyhow!("can't provide type without parent"));
+                }
+                if d.mdev_type.is_none() || d.parent.is_none() {
+                    return Err(anyhow!("Device is insufficiently specified"));
+                }
             }
         }
     }
--- a/mdevctl.8
+++ b/mdevctl.8
@@ -148,11 +148,19 @@ are applied the next time the device is started.
 .PP
 \fBstart\fR \fIDEVICESPEC\fR
 .RS 4
-Start an mdev device, identified by its UUID and optionally its parent,
-or its parent and type and optionally its UUID, which is generated if
-not given.
-If specified via its parent and optionally its UUID, the type may be
-specified in a JSON configuration file, alongside additional parameters.
+Start a mediated device. This command can be used to start either a
+previously-defined device or a newly-created transient device.
+
+If the UUID and optional parent argument matches an existing device definition,
+then the existing device will be started. It is an error to specify a device
+type that conflicts with the existing device definition.
+
+If the UUID argument is omitted or if the specified UUID and parent does not
+match an existing device definition, a new transient device will be started.
+If the UUID is omitted, a new UUID will be automatically generated. When
+starting a new transient device, the parent and device type must be specified.
+A \fB--jsonfile\fR may replace the \fB--type\fR specification and also include
+additional attributes in JSON format to be applied to the started device.
 .RE
 
 .PP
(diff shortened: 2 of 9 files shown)


collinwalling:
Looks like this will start a device with `--uuid --type --parent` in a way that makes sense. I would suggest a message on stdout declaring `"device <UUID> started from <path/to/device/json>"` when starting a defined device, or `"device <UUID> created and started from command line"` when starting a new device from scratch. I think the latter message would be the most useful, that way the user knows the UUID of the newly created device (in case they use `uuidgen`).

jonner:
For the record, if the UUID of the device was *not* specified in the start command, mdevctl *will* print UUID that was internally generated. So the user will know the UUID in that case (but won't if they passed a \`uuidgen\` command inline). So I'm a little hesitant to change the output or make it more verbose in case somebody used that UUID command output in a script in the past.

collinwalling:
Okay. At the very least I think the man page should be re-worded for clarity. What if we go from something like:
```
       start DEVICESPEC
           Start  an  mdev  device,  identified by its UUID and optionally its
           parent, or its parent and type and optionally its  UUID,  which  is
           generated if not given.  If specified via its parent and optionally
           its UUID, the type may be specified in a JSON  configuration  file,
           alongside additional parameters.
```

to:

```
       start DEVICESPEC
           Start a defined or new mdev. A defined mdev with a matching UUID and
           optionally a parent and type will be started and will be prioritized
           over creating a new device. A new device may be created with a parent,
           type, and optionally a UUID, which is generated if not given. For new
           devices, if specified via its parent and optionally its UUID, the type
           may be inferred by a JSON configuration file alongside additional
           parameters and attributes.
```
?

jonner:
Sure, that help text is a little bit confusing. I pushed an additional patch with a counter-proposal. Comments welcome.

cohuck:
[Sigh, github does not seem to allow commenting on a commit.]

The documentation change looks good to me.

src/main.rs
```diff
@@ -225,30 +225,49 @@ fn start_command_helper(
...
                 d.parent = parent;
                 d.mdev_type = mdev_type;
                 dev = Some(d);
             }
+
+            if let Some(ref d) = dev {
+                if d.mdev_type.is_some() && d.parent.is_none() {
+                    return Err(anyhow!("can't provide type without parent"));
+                }
+                if d.mdev_type.is_none() || d.parent.is_none() {
```
collinwalling, line 267: Could this be shortened to just `if d.mdev_type.is_none()` ?

src/main.rs
```diff
@@ -225,30 +225,49 @@ fn start_command_helper(
...
-                    };
+            // if the user specified a uuid, check to see if they're referring to a defined device
+            if uuid.is_some() {
+                let devs = defined_devices(env, uuid.as_ref(), parent.as_ref())?;
+                let n = devs.values().flatten().count();
+                if n > 1 {
+                    return Err(anyhow!(
+                        "Multiple definitions found for device {}",
+                        uuid.unwrap().to_hyphenated().to_string()
+                    ));
```
collinwalling, line 236: The only way to could encounter this error is if the user specifies a uuid without a parent and there are multiple parents who have children with the same uuids, right?

What about adding a little more to the message here: "Multiple definitions found for device {}_, please specify parent_"

collinwalling:
LGTM. Do what you think makes sense with with my two comments above. I think the changes proposed here are great. Thanks for taking a look at this, Jonathon.

src/main.rs
jonner: I guess in theory it could be shortened to that due to the preceding `if` block. But I have to think about it a lot more to convince myself of that fact, whereas the current code is more instantly obvious to me: if either one is missing at this point, emit a generic error.

src/main.rs
jonner: that's a nice user-friendly improvement, I think.

PR #43, opened 2021-07-06
Miscellaneous updates

jonner:
Some minor re-factoring of the test framework, some new tests, and a few minor improvements.


cohuck:
Only looked at patches 1, 4, and 5, but those look good to me.

PR #45, opened 2021-07-14
switch tempdir (deprecated) to tempfile

cpaelzer:
Tempdir has been merged with tempfile since tempfile 3.0.
Comparing the two usage codes of the TempDir struct [1] vs [2]
the change appear clear and only needed because mdevctl
uses a prefix (otherwise a change to the use reference would
have been enough).

[1]: https://github.com/rust-lang-deprecated/tempdir
[2]: https://docs.rs/tempfile/3.2.0/tempfile/struct.TempDir.html

Fixes: !44

Signed-off-by: Christian Ehrhardt <christian.ehrhardt@canonical.com>

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -4,7 +4,8 @@ use std::collections::BTreeMap;
 use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
-use tempdir::TempDir;
+use tempfile::Builder;
+use tempfile::TempDir;
 use uuid::Uuid;
 
 use crate::environment::Environment;
@@ -38,7 +39,7 @@ impl Environment for TestEnvironment {
 impl TestEnvironment {
     pub fn new(testname: &str, testcase: &str) -> TestEnvironment {
         let path: PathBuf = [TEST_DATA_DIR, testname].iter().collect();
-        let scratchdir = TempDir::new(format!("mdevctl-{}", testname).as_str()).unwrap();
+        let scratchdir = Builder::new().prefix("mdevctl-test").tempdir().unwrap();
         let test = TestEnvironment {
             datapath: path,
             scratch: scratchdir,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ log = "0.4"
 serde_json = {version = "1.0", features = ["preserve_order"]}
 structopt = "0.3.13"
 uuid = {version = "0.8.2", features = ["v4"]}
-tempdir = "0.3"
+tempfile = "3"
 
 [build-dependencies]
 structopt = "0.3.13"


cpaelzer:
With this in place I get the tests to pass on recent Debian/Ubuntu

```
running 9 tests
test tests::test_invalid_files ... ok
test tests::test_load_json ... ok
test tests::test_define ... ok
test tests::test_stop ... ok
test tests::test_modify ... ok
test tests::test_start ... ok
test tests::test_types ... ok
test tests::test_undefine ... ok
test tests::test_list ... ok

successes:

---- tests::test_invalid_files stdout ----
[2021-07-14T14:29:50Z WARN  mdevctl] Can't determine uuid for file 'invalid-uuid-value'

[2021-07-14T14:29:50Z WARN  mdevctl] Can't determine uuid for file 'invalid-uuid-value'



successes:
    tests::test_define
    tests::test_invalid_files
    tests::test_list
    tests::test_load_json
    tests::test_modify
    tests::test_start
    tests::test_stop
    tests::test_types
    tests::test_undefine

test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
```

src/tests.rs
```diff
@@ -38,7 +39,7 @@ impl Environment for TestEnvironment {
 impl TestEnvironment {
     pub fn new(testname: &str, testcase: &str) -> TestEnvironment {
         let path: PathBuf = [TEST_DATA_DIR, testname].iter().collect();
-        let scratchdir = TempDir::new(format!("mdevctl-{}", testname).as_str()).unwrap();
+        let scratchdir = Builder::new().prefix("mdevctl-{}").tempdir().unwrap();
```
jonner, line 42: This will generate tempdirs with a literal `mdevctl-{}` prefix rather than the name of the test.

In other words: previously it might have generated tempdirs named `/tmp/mdevctl-listM1agmH`, it will now generate a temp dir named `/tmp/mdevctl-{}M1agmH`.

jonner:
Thanks for the patch. I was not aware that tempdir was deprecated. I see no reason not to switch in that case.

src/tests.rs
jonner: For the record, I don't think it's really that important to keep a per-test prefix. We could just shorten it to "mdevctl" or "mdevctl-test". I do like having the prefix though so that the user knows who to blame if a tempdir gets left behind for some reason.

cpaelzer: Thanks for having a loo and catching this.
I was trying to replicate this without fully rust-knowing what I do.
I'll drop the `-{}`and replace it with `mdevtl-test` in my proposed branch.

jonner:
Looks fine. Thanks for the patch!

PR #48, opened 2021-09-10
Add a github workflow for releases

jonner:
Add two new github workflow jobs that run when a new release is made.

In order to make packaging easier for downstream linux distributions,
we want to include a separate 'vendor' tarball alongside the mdevctl
source. In order to publish this tarball, add a new github workflow that
runs `cargo vendor` and uploads the vendor tarball to the appropriate
release page.

A second job uploads the new release to crates.io using a token that has
been added to the repository secrets.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,45 @@
+name: Release tasks
+on:
+  release:
+    types: [published]
+jobs:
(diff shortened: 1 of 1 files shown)


jonner:
FYI: I just ammended the commit to remove the 5 or 6 trailing blank lines from the end of the new file.

PR #55, opened 2021-12-01
Fix sporadic callout test failures

jonner:
Downstream RHEL builds were failing sporadically due to some racy failures in the callout tests. Make the build more deterministic by fixing the race condition.

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -240,8 +240,8 @@ impl Callout {
         for s in dir.as_ref().read_dir().ok()? {
             let path = s.ok()?.path();
 
-            match self.invoke_script(dev, &path, event, action).ok() {
-                Some(res) => {
+            match self.invoke_script(dev, &path, event, action) {
+                Ok(res) => {
                     if res.status.code().is_none() {
                         warn!("callout script {:?} was terminated by a signal", path);
                         continue;
@@ -255,8 +255,8 @@ impl Callout {
                         );
                     }
                 }
-                _ => {
-                    debug!("failed to execute callout script {:?}", path);
+                Err(e) => {
+                    debug!("failed to execute callout script {:?}: {:?}", path, e);
                     continue;
                 }
             }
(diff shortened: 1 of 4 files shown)


jonner:
@collinwalling any thoughts?

collinwalling:
Makes sense to me. Didn't think to chomp away at stdin. Good catch.

cohuck:
Changes make sense to me as well.

collinwalling:
As an aside, I wonder if the "rust script template" project I was working on would be beneficial to testing in anyway? I don't know what the test cases look like outside of the simple bash scripts I provided, but the template aims to simplify implementing functions in rust for handling the callouts (handles parameter processing and provides a shell for each pre/post-command).

@jonner if I may ask, have you been able to leverage the callouts API in anyway? I recall that you were working on something with libvirt and nodedevs. Would my template be of any use?

I haven't touched it in a while, but I'd certainly love an excuse to revisit. Here's a link: https://github.com/collinwalling/mdevctl_script_tests

jonner:
It could potentially be beneficial. I guess I'm kind of waiting to see how the callouts API gets used before putting a lot more effort into additional testing related to callouts. Speaking of which, unfortunately I haven't yet had time to do anything with libvirt nodedevs and the callout API.

collinwalling:
> It could potentially be beneficial.

Good enough for me :)

> I guess I'm kind of waiting to see how the callouts API gets used before putting a lot more effort into additional testing related to callouts. 

That's fair. I imagine any hidden bugs will be made apparent as other devs start to take advantage of the callouts (we found some during internal testing as this feature was developed).

> Speaking of which, unfortunately I haven't yet had time to do anything with libvirt nodedevs and the callout API.

Also fair! Was simply curious. Feel free to CC me on any upstream conversations when you revisit this in the future, just in case you'd like for someone to chime in with respect to the callouts API.

PR #57, opened 2022-01-05
Handle some config filesystem errors

jonner:
When there are missing dirs or dirs with insufficient permissions, we weren't providing useful error messages. Make these a little easier to debug.

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -1,5 +1,7 @@
 //! A filesystem environment for mdevctl
 
+use anyhow::{anyhow, Result};
+use log::debug;
 use std::path::{Path, PathBuf};
 
 /// A trait which provides filesystem paths for certain system resources.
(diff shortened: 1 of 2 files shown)


cohuck:
The improved error messages are certainly a win.

I'm wondering whether we should consider /etc/mdevctl.d/ a mandatory directory, i.e. checking for it even before looking at any commands. This would tell packagers very early on that something is broken (simple invocation of mdevctl). The downside is that we would check for the directory every time, even if the user simply wanted the help output.

I would be fine with your patch as well.

jonner:
OK, new branch pushed with your suggestion to do a sanity check at startup. I seem to remember another packaging error recently where we forgot to create the callout script dirs in the package, which would have been caught by this patch as well. Perhaps it's too draconian to refuse to continue if e.g. the callout dirs don't exist though?

src/environment.rs
```diff
@@ -33,6 +35,24 @@ pub trait Environment {
     fn notification_dir(&self) -> PathBuf {
         self.scripts_base().join("notifiers")
     }
+
+    fn self_check(&self) -> Result<()> {
+        debug!("checking that the environment is sane");
+        // ensure required system dirs exist. Generally distro packages or 'make install' should
+        // create these dirs.
+        for dir in vec![
+            self.mdev_base(),
```
cohuck, line 44: This one (and parent_base) are sysfs directories, which will not exist unless the mdev kernel driver is loaded. I would say that it is perfectly valid to load mdev-related modules only when devices that can actually be used with it are present. Pre-configuring things in mdevctl and hotplugging devices later should be possible.

I'm not 100% sure whether requiring the callout and notification directories is strictly needed; OTOH, they should be created by the same setup process that also creates /etc/mdevctl.d/, so I think we can as well require all three of them.

jonner: Good point. I will drop those sysfs dirs from the sanity check.

cohuck:
Thanks, this looks good to me now.

PR #58, opened 2022-02-13
fix build.rs to allow specify exact path or name of the rst2man

TheAifam5:
Some distros (like Gentoo) does not have a "rst2man" but "rst2man.py". This PR allows to specify the path of the rst2man or different name of it.


jonner (commented):
Thanks so much for the patch. In addition to the other minor comments, it might be nice to modify the error message slightly to suggest setting the env variable if 'rst2man' is not found? Perhaps something like:

```Unable to generate manpage. Is 'rst2man' installed? You can specify a custom 'rst2man' executable by setting the RST2MAN environment variable.```

If you'd like, I can fix the minor issues and commit. Or you can resubmit if you prefer.

PR #61, opened 2022-05-04
Provide useful error message when parent capitalization is wrong

### LINKED ISSUE #60: `mdevctl` doesn't ignore case sensitivity for `--parent`
The `--parent` argument is used to specify a PCI device in the BDF notation. In my case, it's an NVIDIA GPU on ` 0000:2b:00.0`. 

```mdevctl start --uuid `uuidgen` --type nvidia-46 --parent 0000:2b:00.0``` works, but ```mdevctl start --uuid `uuidgen` --type nvidia-46 --parent 0000:2B:00.0``` returns `Error: No available instances of nvidia-46 on 0000:2b:00.0` because `mdevctl` checks for `/sys/bus/pci/devices/<parent>/mdev_supported_types` to be a directory, but the directory represented by `<parent>` is case sensitive (must be lowercase). This nuance in `--parent` is very subtle and difficult to spot, as the BDF numbers are hexadecimal and therefore the chars could be represented in either caps or lowercase.

jonner:
Proposal to resolve #60

### DIFF
--- a/mdevctl.rst
+++ b/mdevctl.rst
@@ -63,7 +63,8 @@ The following options are understood:
     for the ``modify`` command.
 
 ``-p|--parent=PARENT``
-    Specify or identify the device by its parent device.
+    Specify or identify the device by its parent device. Note that the parent
+    device is specified by its kernel sysfs name and is case-sensitive.
 
 ``-t|--type=TYPE``
     Specify or identify the device by its type.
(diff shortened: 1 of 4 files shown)


cohuck:
Looks good; do we also want to tweak the documentation, stating that --parent takes a case sensitive parameter?

jonner:
latest push adds a small tweak to the manpage for --parent

cohuck:
> latest push adds a small tweak to the manpage for --parent

Looks good to me.

PR #62, opened 2022-06-09
Port CLI to clap v3

jonner:
Clap v3 has been out for some time now and it is intended as a successor
to the structop/clapv2. Port to this new API.


cohuck:
I didn't follow this down into all details, but looking at the clap v3 changelog this conversion seems fine to me.

PR #68, opened 2022-09-01
Callout ordering and cleanups

jonner:
Make sure that we try callout scripts in a predictable order.

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -233,9 +233,6 @@ impl Callout {
             action,
             dev.mdev_type.as_ref()?
         );
-        if dir.as_ref().read_dir().ok()?.count() == 0 {
-            return None;
-        }
 
         for s in dir.as_ref().read_dir().ok()? {
             let path = s.ok()?.path();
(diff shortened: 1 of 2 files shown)


src/callouts.rs
```diff
@@ -234,9 +234,12 @@ impl Callout {
...
         );
 
-        for s in dir.as_ref().read_dir().ok()? {
-            let path = s.ok()?.path();
-
+        for path in dir
+            .as_ref()
+            .read_dir()
+            .ok()?
+            .filter_map(|k| k.ok().map(|e| e.path()))
```
jonner, line 241: Actually, this changes behavior a little bit. It skips error results (e.g. permissions issues, perhaps) rather than reporting the error back to the caller...

jonner: On second (third?) thought, this behavior change seems like an improvement. Previously, if we received an IO error while iterating the list of paths, we would immediately return `None`. This indicated to the calling function that no callout script was found for this device. But perhaps one of the later scripts would have matched this device. So now continue on in the face of IO errors and try the rest of the scripts to see if one is found.

PR #69, opened 2022-09-20
Change callout dir location to /usr/lib

jonner:
New default location for callout scripts is in /usr/lib/, but keep the /etc
location for backwards compatibility. Scripts in /usr/lib have priority over
/etc

- Add system callout script dir outside /etc.
- Add tests for callout dir priority


mdevctl.rst
```diff
@@ -523,11 +523,11 @@ FILES
 Configuration files are in one subdirectory per parent device and named
 by UUID.
 
-*/etc/mdevctl.d/scripts.d/callouts/**
```
cohuck, line 526: I think we need to keep mentioning the old directories (with "DEPRECATED" appended?)

While nothing new should be put there, we don't want people to be surprised when stale scripts from the old directories end up being invoked.

jonner:
Changed the man page to mention the deprecated old location and also added changes to the rpm spec file and makefile that I had missed before.

cohuck:
LGTM

PR #70, opened 2022-11-14
mdevctl.rst: align use of term state

fiuczy:
Use the term state in sections EVENT SCRIPTS, CALL-OUT EVENT SCRIPTS and NOTIFICATION EVENT SCRIPTS.


jonner:
Looks good, thanks.

PR #71, opened 2022-11-14
Active attributes

fiuczy:


src/main.rs
```diff
@@ -513,12 +513,15 @@ fn list_command_helper(
...
-                        if let Ok(attrs) = Callout::get_attributes(&mut dev) {
+                    if let Ok(attrs) = Callout::get_attributes(&mut dev) {
+                        if attrs.is_null() {
+                            // fallback: old behavior - fetch definition
+                            let _ = dev.load_definition();
+                        } else {
                             let _ = dev.add_attributes(&attrs);
                         }
+                    } else {
+                        let _ = dev.load_definition();
```
jonner, line 524: Looking back at the commit history, it appears that one of the reasons that this `dev.load_definition()` call is in the code here is to query the autostart status for the device (in addition to attributes). So I don't think that simply omitting the call to `load_definition()` is the right approach here.

fiuczy: I updated the pull request. Now the code first runs load_definition() and than updates the attributes.

jonner: so... I just went back and looked at the old shell script version of mdevctl, and it does not actually show the autostart status or attributes when listing active devices. That's a behavior difference between the new and old code.

In our help text for `mdevctl list`, we explicitly state that we won't return attributes for active devices:

```
$ mdevctl list --help
mdevctl-list 1.2.0
List mediated devices

With no options, information about the currently running mediated devices is provided. Specifying
'defined' lists the configuration of defined devices, regardless of their running state. This may be
further reduced by specifying specific 'uuid' or 'parent' devices to list. The 'dumpjson' option
provides output listing in machine readable JSON format. When a 'uuid' option is provided and the
result is a single device, the output contains only the JSON fields necessary to recreate a config
file for the device (minus attributes for listings of running devices). When the verbose option is
provided, the human readable listing will include attributes for the device(s).
```

src/main.rs
```diff
@@ -515,8 +515,9 @@ fn list_command_helper(
 
                     let _ = dev.load_definition();
 
-                    if !dev.is_defined() {
-                        if let Ok(attrs) = Callout::get_attributes(&mut dev) {
+                    if let Ok(attrs) = Callout::get_attributes(&mut dev) {
+                        if !attrs.is_null() {
+                            dev.remove_all_attributes();
```
jonner, line 520: `dev.attrs` is actually public, so you could simply call dev.attrs.clear() directly rather than adding a new `remove_all_attributes()` function.

src/main.rs
fiuczy: The code does not follow that documentation. To me the documentation is actually trying to outline the fact that in case a definition does not exist mdevctl itself cannot retrieve the attributes itself.
There are two scenarios:
  a) mdev definition exist; mdev is started/active
  b) mdev definition not exist; mdev is started/active

Let's have a look at mdevctl v1.2.0 called with "mdevctl list --dumpjson".
Scenario 1) Without a callout script:
  In a) the definition is returned including all attributes from the stored definition.
  In b) the definition is returned without attributes.
Scenario 2) With a callout script capable to return the attributes:
  In a) the definition is returned including all attributes from the stored definition.
  In b) the definition is returned with attributes provided by the callout script.
 
What do the two patches now change
Patch 1:
  Changes scenario 1) a) to return the definition is returned without attributes.
  Changes scenario 2) a) to return the definition with attributes provided by the callout script.
  This essentially merges a) and b) into one behavior per scenario 1) and 2).
Patch 2 (which I marked as optional in the commit message) squashed into patch 1:
  Does not change scenario 1).
  Changes scenario 2) a) to return the definition with attributes provided by the callout script.

So in my opinion with patch 1 and without patch 2 the behavior does match the documentation more than what it currently does. If my assumption about what the documentation is actually trying to outline (see above) it would be an exact match.

src/main.rs
fiuczy: True. I thought manipulate the MDev "internals" from outside would not be a good idea.
I have no problem if you think the additional function is to much overhead to cover "internals".

src/main.rs
jonner: yeah, I've changed my mind. I think that when listing active devices, we should not include the attributes from the persistent definition. Returning these attributes might even be misleading since the definition may have been changed since the mdev was started. So I think that we should match the behavior of the old shell-script mdevctl. In other words, I think we should drop the second patch. 

I'll try to push a couple additional commits to your branch with my suggestions.

src/main.rs
```diff
@@ -515,11 +515,14 @@ fn list_command_helper(
 
                     let _ = dev.load_definition();
 
+                    // don't show attributes from the persistent definition if we're listing
+                    // active mdevs. The definition may have changed since the mdev was started
+                    dev.attrs.clear();
```
jonner, line 520: Here I'm clearing attrs even if there is no callout script. This matches the old behavior better and avoids returning possibly-inconsistent information.

src/main.rs
```diff
@@ -515,11 +515,14 @@ fn list_command_helper(
...
+
+                    // if the device is supported by a callout script that gets attributes, show
+                    // those in the output
                     if let Ok(attrs) = Callout::get_attributes(&mut dev) {
-                        dev.remove_all_attributes();
-                        if !attrs.is_null() {
-                            let _ = dev.add_attributes(&attrs);
-                        }
+                        let _ = dev.add_attributes(&attrs);
```
jonner, line 525: `dev.add_attributes()` will already handle the case where `attrs.is_null()` is true.

src/main.rs
fiuczy: Looks fine

src/main.rs
fiuczy: ok, fine.

src/main.rs
fiuczy: Your additional changes look fine and the testing I did with it also turned out OK.

PR #73, opened 2022-11-25
modify: add jsonfile optional parameter

fiuczy:
To allow updates of multiple attributes with a single mdevctl call adding the jsonfile optional parameter to command modify.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -97,7 +97,8 @@ pub enum MdevctlCommands {
                 format that is accepted by the attribute. Upon device start, mdevctl will go \
                 through each attribute in order, writing the value into the corresponding sysfs \
                 attribute for the device. The startup mode of the device can also be selected, auto \
-                or manual. \n\n\
+                or manual. Alternatively, the 'jsonfile' option may be used to replace the startup \
+                mode and any attributes with the contents of the specified file.\n\n\
                 Running devices are unaffected by this command."
     )]
     Modify {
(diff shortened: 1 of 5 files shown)


src/cli.rs
```diff
@@ -97,7 +97,8 @@ pub enum MdevctlCommands {
                 format that is accepted by the attribute. Upon device start, mdevctl will go \
                 through each attribute in order, writing the value into the corresponding sysfs \
                 attribute for the device. The startup mode of the device can also be selected, auto \
-                or manual. \n\n\
+                or manual. If modified via 'jsonfile', then 'startup', and any attributes are provided \
```
cohuck, line 100: I found this sentence a bit confusing... IIUC, 'jsonfile' conflicts with the other method?

Maybe something like

```

Alternatively, the 'jsonfile' option may be used to replace the startup mode and any attributes with the contents of the specified file.
```

fiuczy: I borrowed the wording from Define but have to agree that yours reads better.

cohuck: Oh, indeed... we may want to tweak the wording there as well, but as nobody complained before, we can also keep it as-is.

fiuczy: I will push an rebased version including your suggested rewording.

cohuck:
Looks good to me.

cohuck (approved):

jonner:
Looks fine to me as well.

Issue #72, opened 2022-11-15
modify command: add CLI option to modify stored definition or started mdev only


fiuczy:
When a definition is started currently the modifications are only applied to the definition and not to the started mdev.
If a started mdev instance supports modifications these either could be done manually in the sysfs or with a callout script in mdevctl.

Idea:
Add an optional parameter to run a modify command on either the definition or started mdev instance only.

Please consider the following scenarios:
If the started mdev instance does not support a modify or there is no callout script that supports modifying a started mdev a parameter to modify a started mdev instance only would be a no-op without a (error?) message or any other indication to the user.
Adding an option for modifying the definition only is essentially what mdevctl does today without any parameter and to prevent an update of the started mdev instance this information has be provided to the callout script or the callout script must not be called at all.

fiuczy:
Here is an idea to make a live option fit in with mdevctl much better:
Extend the external script interface with a new event "live".

When the "live" CLI option is provided on the modify command, mdevctl checks if the mdev is running.
- If it is not running the modify command returns with an error message "error: mdev is not running".
- If it is running the modify command creates a callout with the new event "live".
	- If no callout script is found responsible the modify command returns with an error message "error: mdev live update is not supported".
	- If a callout script satisfies the request it can succeed or fail providing the appropriate return code and message.

Existing callout scripts should not be impacted by the new event and mdevctl would remain in control of the updates.

fiuczy:
All of this was resolved with PR #108 Callout version live v6 as well and therefore I am closing this issue.

PR #76, opened 2022-12-13
Fix RC handling in callout_dir

fiuczy:
If callout script is set the special return code 2 which represent NoMatchingScript found needs to be added to the rc matching.

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -343,6 +343,7 @@ impl Callout {
 
         match rc {
             Some(0) => Ok(()),
+            Some(2) => Err(CalloutError::NoMatchingScript),
             Some(n) => Err(CalloutError::InvocationFailure(
                 self.script.as_ref().unwrap().to_path_buf(),
                 Some(n),


cohuck:
Doesn't invoke_first_matching_script already handle rc 2 correctly? What am I missing?

fiuczy:
> Doesn't invoke_first_matching_script already handle rc 2 correctly? What am I missing?

hmmm... I guess the match for script in line 393?
If script is set in callout than invoke_first_matching_script is NOT call but an immediate invoke_script instead.
In that case rc 2 is not special cased to return None.

cohuck:
Ok, that makes sense. Change lgtm.

jonner:
if `self.script` is `Some`, that means that the `Pre` event found a matching script (e.g. a script that did not return response code 2), and that script got stored in `self.script`. So I would definitely not expect the same script to return a response code of 2 for the `Post` event. Is that what is happening here? Or is there a scenario I'm not thinking of?

fiuczy:
> Or is there a scenario I'm not thinking of?

If the expectation is that only post runs after pre there is no current existing scenario.
My understanding is that in general RC = 2 means the script called does not support the event for the type.
Therefore as the event orchestration is actually done in the method invoke I would expect the decision making to happen in that code instead. 
Btw I hit this while including a new live event which is called between pre event and post event.
I felt this is kind of an independent patch but I can also include this patch also into my live event series.

jonner:
> My understanding is that in general RC = 2 means the script called does not support the event for the type.

That's not my understanding. My understanding is that RC=2 means that the script does not support this device type. If you simply don't support the event, I would expect the script to simply do nothing and return a success response code.

rosatomj:
> > My understanding is that in general RC = 2 means the script called does not support the event for the type.
> 
> That's not my understanding. My understanding is that RC=2 means that the script does not support this device type. If you simply don't support the event, I would expect the script to simply do nothing and return a success response code.


Jonathon's explanation matches my expectation and is exactly what the vfio-ap callout script does today.

AFAIU RC=2 was specifically intended to mean 'This script does not support the mdev type specified'  -- with the idea being that mdevctl could use this response to recognize that it needed to keep scanning scripts.d/callouts/ to try and find a callout script that supports the specified type (or give up and not use callouts for the command).

As far as unknown events or actions, the vfio-ap callout today ignores them and returns 0.

fiuczy:
> > > My understanding is that in general RC = 2 means the script called does not support the event for the type.
> > 
> > 
> > That's not my understanding. My understanding is that RC=2 means that the script does not support this device type. If you simply don't support the event, I would expect the script to simply do nothing and return a success response code.
> 
> Jonathon's explanation matches my expectation and is exactly what the vfio-ap callout script does today.
> 
> AFAIU RC=2 was specifically intended to mean 'This script does not support the mdev type specified' -- with the idea being that mdevctl could use this response to recognize that it needed to keep scanning scripts.d/callouts/ to try and find a callout script that supports the specified type (or give up and not use callouts for the command).
> 
> As far as unknown events or actions, the vfio-ap callout today ignores them and returns 0.

Seems like it is time to think about a capabilities callout for the specified type and build upon that instead of relying on RCs.

jonner:
> Seems like it is time to think about a capabilities callout for the specified type and build upon that instead of relying on RCs.

That seems like a fairly significant change to the callouts infrastructure, and I've not seen evidence that it is needed. Can you explain why returning 0 and doing nothing doesn't work for your use case?

fiuczy:
> > Seems like it is time to think about a capabilities callout for the specified type and build upon that instead of relying on RCs.
> 
> That seems like a fairly significant change to the callouts infrastructure, and I've not seen evidence that it is needed. Can you explain why returning 0 and doing nothing doesn't work for your use case?

As I as wrote before I am currently working on a live/defined modify which updates stored configs and active mdevs. Kind of like outlined in https://github.com/mdevctl/mdevctl/issues/72.

jonner:
I still can't imagine why you can't make that work with the approach I outlined. Maybe I'll have to wait to see your patch.

fiuczy:
As PR #108 Callout version live v6 has been merged I think this initial PR can also get closed. Any objections?

jonner:
Fine with me

PR #77, opened 2023-02-22
undefine: report error if device is not undefined

fiuczy:
If a callout script returns with an error code the undefine method is not called and therefore the mdev is not undefined. The user should get notified this occured instead of silently suppressing the error when the mdev is not undefined.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -158,14 +158,28 @@ fn define_command(
 /// Implementation of the `mdevctl undefine` command
 fn undefine_command(env: &dyn Environment, uuid: Uuid, parent: Option<String>) -> Result<()> {
     debug!("Undefining mdev {:?}", uuid);
+    let mut failed = false;
     let devs = defined_devices(env, Some(&uuid), parent.as_ref())?;
     if devs.is_empty() {
         return Err(anyhow!("No devices match the specified uuid"));
     }
     for (_, mut children) in devs {
         for child in children.iter_mut() {
-            let _ = Callout::invoke(child, Action::Undefine, |dev| dev.undefine());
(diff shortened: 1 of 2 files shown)


jonner:
Out of curiosity, what are some valid use cases for the `undefine` callout script to fail? 

If the callout script fails, there's nothing a user can do to 'correct' the error and successfully undefine the device; the `uuid` and `parent` values are validated before we even get to the callout. So I wonder if we should even let the `undefine` callout script block primary command execution. I worry a little bit about a user being prevented from undefining any devices due to a badly behaved callout script. Although maybe a better solution to that is adding a `--force` argument that forces the primary command to be executed regardless of callout script results...

fiuczy:
The current code allows the pre callout to stop/prevent the `undefine`. The patch just tries to give the user awareness that this happened. I also agree with you that an new `--force` option to override the callout scripts voting would be the best way to cope with this but it would be another patch.

Thinking about a reason why a script prevents a defined config from being undefined could e.g. that the defined config has a running instance.

jonner:
There should be no problem undefining a device that has a running instance. But I'll merge this patch and maybe I'll add a new --force option in a later patch.

fiuczy:
Yes, there can be running instances without a definition in mdevctl.
But there could also be a theoretical operational model in which a defined configuration should exist if it is managed via mdevctl and therefore the callout script managing that mdev type prevents an undefine of the defined configuration as long as the matching instance is running. I am not saying I want to do something like this but asked for a valid use case. :)

fiuczy:
> There should be no problem undefining a device that has a running instance. But I'll merge this patch and maybe I'll add a new --force option in a later patch.

Just so we don't do the work twice: I will send a patch today.

PR #79, opened 2023-03-03
undefine: add force option

fiuczy:
The force option on command undefine allows to override the decision of a callout script to reject in the callout pre event the undefine.

Reviewed-by: Steffen Eiden <seiden@linux.ibm.com>

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -114,6 +114,15 @@ impl Display for State {
     }
 }
 
+impl<T, E> From<&Result<T, E>> for State {
+    fn from(value: &Result<T, E>) -> Self {
+        match value {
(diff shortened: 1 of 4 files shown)


jonner:
Unfortunately we ended up doubling work after all since I created a patch last week but forgot to open a PR for it before I left for a few days off :/  Mine is a bit more ambitious as it adds a 'force' option for all commands that can fail due to callouts: `define`, `undefine`, `modify`, `start`, and `stop`.  I'll push mine anyway and we can compare them.

jonner:
See PR #80

jonner:
I merged #80 instead as discussed.

PR #80, opened 2023-03-07
Add 'force' option for commands that can fail due to callouts

jonner:
- Add --force option for commands that have callouts

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -127,26 +127,38 @@ impl Callout {
         }
     }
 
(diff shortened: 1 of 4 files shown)


fiuczy:
Hi Jonathon,
looks good to me: Reviewed-by: Boris Fiuczynski <fiuczy@linux.ibm.com>

I also gave a real tryout and it worked for me.
Therefore let's simply take your implementation.

jonner:
OK, thanks for testing.

PR #83, opened 2023-04-04
Callout version live v2

fiuczy:
This series implements versioning in the callout script API. The versioning allows to add the new callout event `live` which allows callout scripts to support live modifications of an started mdev device.

### DIFF
--- a/mdevctl.rst
+++ b/mdevctl.rst
@@ -41,7 +41,8 @@ The following options are understood:
 
 ``-d|--defined``
     List all defined devices, even if not active. Valid for the ``list``
-    command.
+    command. Modify the defined configuration of a device, even if the
+    device is active. Valid for the ``modify`` command.
 
 ``--delattr``
     Delete an attribute entry. Valid for the ``modify`` command.
@@ -57,6 +58,10 @@ The following options are understood:
 ``--jsonfile=FILE``
     Read the configuration for a device from a JSON file *FILE*.
     Valid for the ``define`` and ``start`` commands.
+
+``-l|--live``
+    Modify active device without modifying the defined configuration of
+    the device. Valid for the ``modify`` command.
 
 ``-m|--manual``
     Do not start a device automatically on parent availability. Valid
@@ -111,8 +116,12 @@ The following commands are understood:
     Attributes can be added or deleted. Attributes to be deleted must be
     specified by their index; if an attribute is specified without an
     index, it is appended at the end of the attribute list.
-    Running devices are unaffected by this command; changes in the configuration
-    are applied the next time the device is started.
+    Active devices are unaffected by this command; changes in the configuration
+    are applied the next time the device is started. Depending on installed
+    callout scripts active devices can be modified. With ``-l|--live``
+    modifications can be applied to active devices if a callout scripts supports
+    the event ``live``. The option ``-d|--defined`` also direct the modification
+    to the started device configuration.
 
 ``start`` *DEVICESPEC*
     Start a mediated device. This command can be used to start either a
@@ -353,6 +362,7 @@ Essentially, the procedure in mdevctl looks like this:
 
     - command-line parsing & setup
     - invoke pre-command call-out
+    - invoke live-command call-out
     - primary command execution [1]_
     - invoke post-command call-out [1]_
     - invoke notifier
@@ -378,7 +388,7 @@ progress, and the mediated device. The parameters are as follows:
 
 ``-e=``\ *event*
     Event type of call-out that is invoked. For call-out scripts, this may be
-    ``pre``, ``post``, or ``get``. For notification scripts, this will
+    ``pre``, ``live``, ``post``, or ``get``. For notification scripts, this will
     always be ``notify``.
 
 ``-a=``\ *action*
@@ -399,8 +409,8 @@ progress, and the mediated device. The parameters are as follows:
 CALL-OUT EVENT SCRIPTS
 ----------------------
 
-A call-out event script is invoked during a ``pre``, ``post``, or ``get``
-event. mdevctl will attempt each script stored in the mdevctl callouts
+A call-out event script is invoked during a ``live``, ``pre``, ``post`` or
+``get`` event. mdevctl will attempt each script stored in the mdevctl callouts
 directory until either a script that satisfies the device type is found or all
 scripts have been attempted. A device script must check the "TYPE" parameter to
 ensure the specified device type is supported, otherwise error code 2 should be
@@ -408,8 +418,25 @@ returned. If no script is found for the specified device type, then mdevctl
 will carry on as normal.
 
 These scripts are stored in */usr/lib/mdevctl/scripts.d/callouts*. The same
-script is invoked for ``pre``, ``post``, and ``get`` call-out events for
-the device type.
+script is invoked for ``live``, ``pre``, ``post``, and ``get`` call-out events
+for the device type.
+
+``Live-Command``
+
+    A live-command call-out event is invoked once before the pre-command call-out
+    event execution. This only occurs if the ``live`` option is specified on the
+    ``modify`` command and the device modified is active.
+    Event type is ``live``. State will always be ``none``.
+
+    If the ``live`` command line option is specified any non-zero return code results in
+    a live modification failure except for all call-outs return with return code 2
+    resulting in a ``live update not supported`` information.
+    The return code is disruptive if also the option ``defined`` is provided and will
+    prevent the update of the defined device configuration.  
+
+    A notification event will follow if the ``live`` command line option is specified.
+
+    This event is only supported for the ``modify`` command.
 
 ``Pre-Command``
 
--- a/mdevctl.rst
+++ b/mdevctl.rst
@@ -361,8 +361,8 @@ file for define, or activating a device for start).
 Essentially, the procedure in mdevctl looks like this:
 
     - command-line parsing & setup
+    - invoke live-command call-out (if live is requested)
     - invoke pre-command call-out
-    - invoke live-command call-out
     - primary command execution [1]_
     - invoke post-command call-out [1]_
     - invoke notifier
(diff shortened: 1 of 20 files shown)


collinwalling:
I can concur. These test succeed locally on my end.

The test that fails on the CI is a get-attributes after a live-modify. I wonder if there's something in the CI that's preventing a live device from being altered?

collinwalling:
Tests pass on Fedora, but fail on Ubuntu. Ubuntu uses `grep -z` which adds a trailing NULL to the string. serde_json does not like this.

Adding a bit of string sanitation to the callouts code that invokes serde_json can fix this. I trimmed the trailing NULL byte and the test passes. I'll propose a brief patch to amend this if it's acceptable.

src/callouts.rs
```diff
@@ -106,6 +196,235 @@ impl Display for State {
...
+}
+
+impl CalloutScripts {
+    pub const fn new() -> Self {
+        CalloutScripts {
+            callouts: Vec::new(),
+        }
+    }
+
+    fn find_script(&mut self, dev: &mut MDev) -> Result<CalloutScript, CalloutError> {
```
jonner, line 302: No need for dev to be `mut` with some changes below

src/callouts.rs
```diff
@@ -106,6 +196,235 @@ impl Display for State {
...
+            "Lookup failed starting to search for mdev type '{:?}'",
+            mdev_type
+        );
+        // search directories for mdev_type parent tuple
+        for dir in dev.env.callout_dirs() {
+            debug!(
+                "Searching in directory {:?} for mdev type '{:?}'",
+                mdev_type, dir
+            );
+            match self.callout_dir_search(&mut dev.clone(), dir.clone()) {
```
jonner, line 325: no need to clone these args with some changes below...

src/callouts.rs
```diff
@@ -106,6 +196,235 @@ impl Display for State {
...
+    }
+
+    fn callout_dir_search(
+        &mut self,
+        dev: &mut MDev,
+        dir: PathBuf,
+    ) -> Result<CalloutScript, CalloutError> {
+        if !dir.is_dir() {
+            return Err(CalloutError::NoMatchingScript);
+        }
```
jonner, line 359: To avoid cloning above:

```suggestion
    fn callout_dir_search<P: AsRef<Path> + std::fmt::Debug>(
        &mut self,
        dev: &MDev,
        dir: P,
    ) -> Result<CalloutScript, CalloutError> {
        if !dir.as_ref().is_dir() {
            return Err(CalloutError::NoMatchingScript);
        }
```

src/callouts.rs
```diff
@@ -106,6 +196,235 @@ impl Display for State {
...
+            match serde_json::from_str(&stdout).map_err(CalloutError::InvalidJSON) {
+                Ok(ce) => ce,
+                Err(e) => return Err(e),
+            };
+        ce.supports.ok_or(CalloutError::NoSupportedVersion)
+    }
+
+    fn invoke_script_capability<P: AsRef<Path> + std::fmt::Debug>(
+        &self,
+        dev: &mut MDev,
```
jonner, line 382:
```suggestion
        dev: &Mdev
```

src/callouts.rs
```diff
@@ -106,6 +196,235 @@ impl Display for State {
...
+        dir: P,
+    ) -> Option<(PathBuf, CalloutVersion)> {
+        let event: Event = Event::Get;
```
jonner, line 385: There's a fair bit of this function that's pretty similar to `Callout::invoke_first_matching_script()`, both in duplicated code and similar functionality. I don't really like this duplication. Both of the functions iterate through the callout directories and test them one by one and return a single matching callout script. I would really like to try to combine them somehow.

src/callouts.rs
```diff
@@ -106,6 +196,235 @@ impl Display for State {
...
+}
+
+impl AsRef<Path> for CalloutScript {
+    fn as_ref(&self) -> &Path {
+        &self.path
+    }
+}
+
+pub struct CalloutScripts {
+    callouts: Vec<CalloutScript>,
```
jonner, line 274: as far as I can tell, this `Vec` only ever should hold a single script. The first one that is found gets pushed into the vector and from then on, this script gets used.

mdevctl.rst
```diff
@@ -468,6 +468,72 @@ the device type.
             }
         ]
 
+``Get-capabilities``
```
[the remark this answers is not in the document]
collinwalling: There could be more applications to this sort of command as well, such as improving the handshaking between mdevctl and the scripts. Right now to tell if a script corresponds to a particular device, mdevctl goes through a trial-and-error of each file in the callouts directory until it finds the right one.

It could be useful to have mdevctl instead query **each** script in the directory, and the scripts can return meta data that includes:
- device type supported
- actions supported
- events supported
- other data that may be useful

mdevctl can retain this info in a struct of the parsed response, as well as a path to the script. The callouts code will no longer need to rely on pinging the script for changes, but rather check the (temporarily) cached data. Of course, fallback to the current method of how things are handled if the query is not supported by the script.

This would also lead the way for the potential of (permanently) caching script meta data, forgoing the directory scan in the future in favor of checking a master file (or maybe some `/path/to/cache/dev_type.cache` for a more direct approach).

src/callouts.rs
collinwalling: I'm not sure if my comment above satisfies your concern here, but hopefully it's somewhere in the correct direction.

mdevctl.rst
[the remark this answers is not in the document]
collinwalling: Hmm, I understand the desire to keep the design simple. However, there's also the desire to modify a running device without a stop/start sequence.

Perhaps some insight as to how notify scripts work would help to understand the API more clearly (I admittedly didn't work closely with those types of scripts during the initial design). IIRC, you worked on a notify script? What would happen when a `notify` event gets invoked after a `modify` action?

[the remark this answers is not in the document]
jonner: No, I didn't actually work on any notify scripts yet. IIRC, we implemented it because it seemed like it would be useful for libvirt (and possibly other things). But by the time the callout stuff was merged, we already had a workaround in libvirt. And I haven't gotten around to switching libvirt to use a notify script.

mdevctl.rst
```diff
@@ -468,6 +468,72 @@ the device type.
...
+
+    A script that wants to support versioning must return a versioning JSON on standard
+    output. The script should list all supported actions in the actions array and all
+    supported events in the events array. It is possible to add additional actions or
+    events in the array but if mdevctl did not have these in the arrays in provides
+    they are ignored. Example::
+
+        {
+          "supports": {
+            "version": "1.1.0",
```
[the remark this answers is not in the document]
fiuczy: Sure, that is no problem as version is just a string.
I agree that a simple integer as version is also much simpler to handle on both ends (script as well as mdevctl).

src/callouts.rs
```diff
@@ -106,6 +196,235 @@ impl Display for State {
...
+}
+
+static CALLOUT_SCRIPTS: Mutex<CalloutScripts> = Mutex::new(CalloutScripts::new());
+
+fn find_callout_script(dev: &mut MDev) -> Result<CalloutScript, CalloutError> {
+    return CALLOUT_SCRIPTS.lock().unwrap().find_script(dev);
+}
```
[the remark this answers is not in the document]
fiuczy: Maybe I misunderstood "I don't like this global singleton pattern.".
Do you dislike global, singleton or both?
Is your suggestion to place the singleton into `Callout `and rename `Callout` into `CalloutManager`?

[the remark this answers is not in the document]
jonner: aha, I was forgetting about the list command where callouts for multiple devices are happening for one command.  In that case a cache could indeed be beneficial. But I still don't love that it's a global variable. We already have `Environment` which is kind of a collection of application-level 'global' data. I wonder if it would fit in there...

mdevctl.rst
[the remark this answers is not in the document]
fiuczy: > I don't really like this name. `get-capabilities` has a very similar naming convention to `get-attributes`, so they seem like they are similar commands. But `get-attributes` queries information about the _device_, whereas `get-capabilities` queries information about the _script_. if we need something like this it almost makes more sense to introduce a new event like `meta`.
> 
> My first impression is that this introduces a lot of additional complexity to the simple callout design we originally had.

There are events and actions in the callout API.
The event `get` fits well with the fact that we want to ask the script about what it supports and therefore I reused it.
I had to add an new action for the querying of the scripts capabilities and selected `capabilities`. I am open to change this to e.g. `probe`, `support` or `supports`. This would result in `get-probe`, `get-support` or `get-supports`.

The event `pre` cannot be reused as I was told that one cannot rely on RCs if the action is unknown. Therefore the combination of event `get` with a new action let's call it for now `capabilities` which a supporting script responses to with a defined output in stdout is an elegant way to handle the problem.

The new workflow is:
Search a script that 
1) supports the mdev type of the mdev device currently in the works and 
2) supports the event `get` and the new action `capabilities`
3) returns with a valid formatted response on stdout
until either 
a) a matching versioning script is found or
b) no matching versioning script is found.
Then execute with whatever event/action mdevctl is supposed to do by first checking if 
A) a versioning script was found and executing it using the old callout pattern of "a-script-has-been-found-already" or 
B) a versioning script was NOT found and use the old callout pattern (using event 'pre' for script search and than use it).

src/callouts.rs
fiuczy: This method has been eliminated when rebasing on your refactoring.

src/callouts.rs
fiuczy: The `Vec` is needed for the case if `mdevctl list` is called and multiple mdevs with different mdev-types exist.

jonner:
see #93

PR #86, opened 2023-04-17
Refactor get attributes

jonner:
I've never been happy with how we handle the `get-attributes` callout action. This series refactors the internals of the callout infrastructure to reduce some duplication and to make it fit better with the other callout actions.

1. First of all, I've refactored the `Callout::callout()` function to make it generic enough (by returning `Output`) so that it can also be used for the `get-attributes` action. This allows us to remove the traversal of callout directories within `Callout::get_attributes()`. 
2. In doing so, I've made it so that if `Callout::script` has already been set (i.e. by a `pre` callout being executed first), `Callout::get_attributes()` will use that script rather than re-iterating the callout script directories to find a script for `get-attributes`. 
3. To actually be able to use the feature mentioned in point 2, we need to change the `func` signature in `Callout::invoke()` to pass a `Callout` object to this callback function. Then we can execute `get_attributes()` from inside of the main operation callback function. This has the additional benefit that the `pre` callout is always issued first before calling `get-attributes`

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -148,6 +148,7 @@ impl Callout {
     {
         let res = self
             .callout(dev, Event::Pre, action)
+            .map(|_output| ()) // can ignore output for general callouts
             .or_else(|e| {
                 force
                     .then(|| {
@@ -342,18 +343,18 @@ impl Callout {
         None
     }
 
-    fn callout(&mut self, dev: &mut MDev, event: Event, action: Action) -> Result<()> {
+    fn callout(&mut self, dev: &mut MDev, event: Event, action: Action) -> Result<Option<Output>> {
         match self.script {
             Some(ref s) => {
                 let output = self.invoke_script(dev, s, event, action)?;
                 self.print_err(&output, s);
                 match output.status.code() {
-                    None | Some(0) => Ok(()),
+                    None | Some(0) => Ok(Some(output)),
                     Some(n) => Err(invocation_failure(self.script.as_ref().unwrap(), Some(n))),
                 }
             }
             None => {
-                let mut res = Ok(());
+                let mut res = Ok(None);
                 for dir in dev.env.callout_dirs() {
                     if !dir.is_dir() {
                         continue;
@@ -363,7 +364,7 @@ impl Callout {
                             self.print_err(&o, &p);
                             self.script = Some(p.clone());
                             match o.status.code() {
-                                Some(0) => Ok(()),
+                                Some(0) => Ok(Some(o)),
                                 Some(n) => Err(invocation_failure(&p, Some(n))),
                                 None => continue,
                             }
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Context, Result};
 use log::{debug, warn};
 use std::ffi::OsStr;
 use std::fmt::{self, Display, Formatter};
@@ -14,6 +14,17 @@ pub enum Event {
     Post,
     Notify,
     Get,
+}
+
+fn invocation_failure(path: &PathBuf, code: Option<i32>) -> anyhow::Error {
+    anyhow!(
+        "Script '{:?}' failed with status '{}'",
+        path,
+        match code {
+            Some(i) => i.to_string(),
+            None => "unknown".to_string(),
+        }
+    )
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


fiuczy:
I am have not looked at how many changes this will create in my versioning/live update series but I will put that aside for now.
Looking at your refactoring I got the following impression:
The `Callout` objects look reusable for any kind of device but they are reusable only for devices of the same mdev type as with the first invocation of a method on a `Callout` object it is tied to the mdev type (without the users knowledge).

My proposal to remove the mdev parameter from the public methods `invoke` and `get_attributes` and add it to `Callout` constructor is not as simple as I first thought as the mdev_type on MDev is optional it does not ensure the mdev type can be tied into the `Callout` object. Looking at the public methods of `Callout` I found that these methods also run into an error far down the call chain if the mdev does not have an mdev type set. So it might be better the cause the the error in the constructor right away. What do you think?

jonner:
> I am have not looked at how many changes this will create in my versioning/live update series but I will put that aside for now. Looking at your refactoring I got the following impression: The `Callout` objects look reusable for any kind of device but they are reusable only for devices of the same mdev type as with the first invocation of a method on a `Callout` object it is tied to the mdev type (without the users knowledge).

That is basically true of the existing design as well. In `Callout::callout()`, after we find a matching script, we store it in the `Callout::script` variable and that gets re-used for the `post` and `notify` events. The difference is that Callout was only allocated internally in `Callout::invoke()` and was freed before returning and now I'm allocating the object in the calling function.

> My proposal to remove the mdev parameter from the public methods `invoke` and `get_attributes` and add it to `Callout` constructor is not as simple as I first thought as the mdev_type on MDev is optional it does not ensure the mdev type can be tied into the `Callout` object. Looking at the public methods of `Callout` I found that these methods also run into an error far down the call chain if the mdev does not have an mdev type set. So it might be better the cause the the error in the constructor right away. What do you think?

I don't know whether that will be feasible or not. Right now we can construct a partial `MDev` object with only a uuid, and then call e.g. `MDev::load_from_json()` to fill in the details from a file. But if we make the mdev_type required at construction this stuff would have to be refactored. If that is possible, and doesn't make the code worse, I'm not opposed to trying it.

fiuczy:
> > I am have not looked at how many changes this will create in my versioning/live update series but I will put that aside for now. Looking at your refactoring I got the following impression: The `Callout` objects look reusable for any kind of device but they are reusable only for devices of the same mdev type as with the first invocation of a method on a `Callout` object it is tied to the mdev type (without the users knowledge).
> 
> That is basically true of the existing design as well. In `Callout::callout()`, after we find a matching script, we store it in the `Callout::script` variable and that gets re-used for the `post` and `notify` events. The difference is that Callout was only allocated internally in `Callout::invoke()` and was freed before returning and now I'm allocating the object in the calling function.

Before the internal use prevented the user of `Callout` from the requirement to know about the mdev type side effect. Now sje/he can use the `Callout` object once and than discard it or know about the mdev type side effect and handle the required restrictions somehow to reuse the object.
I think that this makes things a bit more tricky.

> 
> > My proposal to remove the mdev parameter from the public methods `invoke` and `get_attributes` and add it to `Callout` constructor is not as simple as I first thought as the mdev_type on MDev is optional it does not ensure the mdev type can be tied into the `Callout` object. Looking at the public methods of `Callout` I found that these methods also run into an error far down the call chain if the mdev does not have an mdev type set. So it might be better the cause the the error in the constructor right away. What do you think?
> 
> I don't know whether that will be feasible or not. Right now we can construct a partial `MDev` object with only a uuid, and then call e.g. `MDev::load_from_json()` to fill in the details from a file. But if we make the mdev_type required at construction this stuff would have to be refactored. If that is possible, and doesn't make the code worse, I'm not opposed to trying it.

Actually `Callout` should not have to do any of the `MDev`,lets call it, "preparational setup steps" internal (unless a new `MDev` instance is required). As fare as I can see this is taken care of in `main.rs`. The `MDev` instance(s) specified by the mdevctl user should be "fully specified" at the point in time a callout is triggered. What I found in the public `Callout` methods so far suggests that this is also coded into the public methods as otherwise runtime errors are triggered within `Callout` code.
With this change the `Callout` user would get awareness of the mdev - callout correlation.

Next step could be to have something like a CalloutFactory which takes care of the `Callout` instance management. It could e.g. be the class which holds internally the singleton for mdev type - script lookups.

jonner:
> Before the internal use prevented the user of `Callout` from the requirement to know about the mdev type side effect. Now sje/he can use the `Callout` object once and than discard it or know about the mdev type side effect and handle the required restrictions somehow to reuse the object. I think that this makes things a bit more tricky.

I agree. that is a drawback of my redesign.

> Actually `Callout` should not have to do any of the `MDev`,lets call it, "preparational setup steps" internal (unless a new `MDev` instance is required). As fare as I can see this is taken care of in `main.rs`. The `MDev` instance(s) specified by the mdevctl user should be "fully specified" at the point in time a callout is triggered. 

I see. I think I misunderstood your suggestion. I thought you were suggesting that this should happen in the `MDev` constructor, but you were actually talking about the `Callout` constructor? I don't see any problem with that.

PR #87, opened 2023-05-04
modify: ensure required parameter is given

fiuczy:
Prevent running mdevctl modify command without specifying either parameter addattr, delattr or jsonfile.


jonner:
Thanks for the patch. Looks good!

PR #88, opened 2023-05-04
Refactor get attributes extended

fiuczy:
This series ties together the mdev and the callout objects as once the callout object is used for the first time it finds the callout script specifically for the mdev object provided and every other call on the callout object is with that script found first.
By providing the mdev when creating the callout object and removing it as a parameter from the callout functions this relationship is better expressed.

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Context, Result};
 use log::{debug, warn};
 use std::ffi::OsStr;
 use std::fmt::{self, Display, Formatter};
@@ -14,6 +14,17 @@ pub enum Event {
     Post,
     Notify,
     Get,
+}
+
+fn invocation_failure(path: &PathBuf, code: Option<i32>) -> anyhow::Error {
+    anyhow!(
+        "Script '{:?}' failed with status '{}'",
+        path,
+        match code {
+            Some(i) => i.to_string(),
+            None => "unknown".to_string(),
+        }
+    )
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


src/callouts.rs
```diff
@@ -89,30 +89,35 @@ impl Display for State {
     }
 }
 
-pub struct Callout {
+pub struct Callout<'a> {
     state: State,
     script: Option<PathBuf>,
+    pub dev: MDev<'a>,
```
jonner, line 95: One thing that I really don't like about this approach is that it copies the `MDev` object into the Callout struct. This means that any changes made due to the callout are not reflected in the original device object. That makes things pretty prone to accidental misuse. For example in this excerpt from `define_command()`:

```
    let mut c = callout(dev);
    c.invoke(Action::Define, force, |c| {
        if c.dev.active {
            let attrs = c.get_attributes()?;
            c.dev.add_attributes(&attrs)?;
        }
        c.dev.define()
    })
```

After this code, I now have two variables in scope: `dev` and `c.dev`. If I now use `dev` here, it won't have any attributes that we just retrieved. I need to remember to use `c.dev` instead. I'd much prefer that `Callout` have a mutable reference to the `MDev` so that the original object gets updated. I whipped up an implementation on top of your branch: cd316f8f6fe4339f060238d97c4f133b147d4de9

src/callouts.rs
```diff
@@ -89,30 +89,35 @@ impl Display for State {
...
+pub fn callout(dev: MDev) -> Callout {
+    Callout::new(dev)
 }
 
-impl Callout {
-    pub fn new() -> Callout {
+impl<'a> Callout<'a> {
+    pub fn new(dev: MDev<'a>) -> Callout<'a> {
+        if dev.mdev_type.is_none() {
+            panic!("Device dev must have a defined mdev_type!")
```
jonner, line 105: It seems pretty drastic to `panic!` here. On the other hand, changing the return value to e.g. `Result<Callout>` makes this a little less convenient to use. I'm sort of undecided...

jonner (commented):
Aside from the two specific comments, I generally like it.

src/callouts.rs
fiuczy: Your proposal looks fine to me and I agree that the code in main using the Callout objects is cleaner that way.

src/callouts.rs
fiuczy: I think that the panic is actually not something a user of mdevctl is going to encounter but it is meant to be helpful for detecting the not allowed scenario during running tests when e.g. implementing new features or changing code which results in mdev objects used without the mdev_type being set properly to create a Callout object.
At the same time as you already mentioned I didn't want to sacrifice the easy of use to an "internal programming error".

jonner: yep, I think I've come around to being fine with panic here as it really is a programming error.

src/callouts.rs
jonner: Would you prefer to squash my patch into yours or just keep them separate?

jonner: sounds good

src/callouts.rs
```diff
@@ -370,16 +360,32 @@ impl Callout {
...
-        Ok(())
+                    if !matches!(r, Err(CalloutError::NoMatchingScript)) {
+                        res = r;
+                        break;
+                    }
+                }
+                res
+            }
+        };
+        return res.map_err(anyhow::Error::from);
```
fiuczy: clippy suggests: remove `return`

fiuczy:
I pushed an update with sign-offs, reviewed-bys and your patch squashed in.
Hopefully it meets your expectation and you can merge the PR with master. Thanks

jonner (commented):
In commit d236c28, the commit message has a minor typo: "At the end end of an operation"

jonner:
Also, commit aa94db8 has a commit message typo (probably my fault): "description go write" should be "description to write"

jonner:
> I pushed an update with sign-offs, reviewed-bys and your patch squashed in. Hopefully it meets your expectation and you can merge the PR with master. Thanks

Other than the two minor commit message issues mentioned above, It looks fine. I can either fix the commit messages here or you can force push a new version.

jonner:
I fixed the commit messages and pushed. Thanks!

PR #89, opened 2023-05-04
Give a better error message when user specifies an invalid attribute index

jonner:
Before:
```
$ mdevctl modify --uuid 23c2b183-5876-40f1-90b2-5350b0a03c9b --addattr $NAME --value $VALUE --index=4
Error: Attribute index 4 is invalid
```

After (when a device has attributes):
```
$  mdevctl modify --uuid 23c2b183-5876-40f1-90b2-5350b0a03c9b --addattr $NAME --value $VALUE --index=4
Error: Attribute index 4 is invalid
  Attrs:
    @{0}: {"foo":"bar"}
```

After (when a device has no attributes):
```
$  mdevctl modify --uuid 23c2b183-5876-40f1-90b2-5350b0a03c9b --addattr $NAME --value $VALUE --index=4
Error: Attribute index 4 is invalid
Device 23c2b183-5876-40f1-90b2-5350b0a03c9b has no attributes
```

### DIFF
--- a/src/mdev.rs
+++ b/src/mdev.rs
@@ -239,14 +239,23 @@ impl<'a> MDev<'a> {
         }
 
         output.push('\n');
-        if verbose && !self.attrs.is_empty() {
+        if verbose {
+            let attr_string = self.fmt_attrs();
+            output.push_str(&attr_string);
+        }
+        Ok(output)
+    }
+
+    fn fmt_attrs(&self) -> String {
+        let mut output = String::new();
+        if !self.attrs.is_empty() {
             output.push_str("  Attrs:\n");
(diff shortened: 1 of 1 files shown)


fiuczy (approved):
Looks good to me.

fiuczy (approved):
This comment was made on commit 78df697342f32516c4608700e5d782b1d11530ef
With this change the error message depends on how many attributes the mdev has and if the mdev has many attributes this could get rather lengthy.
Since I guess that it is most likely someone getting the error will run `list -v` anyhow with this change it is no longer necessary to do so.
Therefore I am fine with it. You can add my rb.

PR #91, opened 2023-05-22
modify: fix parameter relations and improve error messages

fiuczy:
Allow mdevctl modify command to use auto and manual parameters without using parameter addattr, delattr or jsonfile. Fixed missing requirement checks and improved error messages.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -118,7 +118,7 @@ pub enum MdevctlCommands {
         group(
             clap::ArgGroup::new("modify")
                 .required(true)
-                .args(&["addattr", "delattr", "jsonfile"]),
+                .args(&["auto", "manual", "addattr", "delattr", "jsonfile"]),
         ),
     )]
     Modify {
@@ -135,7 +135,6 @@ pub enum MdevctlCommands {
         mdev_type: Option<String>,
         #[clap(
             long,
-            conflicts_with("delattr"),
             requires("value"),
             help = "Add a new attribute",
             value_name = "attr_name"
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -141,16 +140,24 @@ pub enum MdevctlCommands {
...
             help = "Value for the attribute specified by --addattr",
             value_name = "attr_value"
         )]
         value: Option<String>,
+        #[clap(long, conflicts_with("addattr"), help = "Delete an attribute")]
+        delattr: bool,
+        #[clap(
+            long,
+            short,
+            conflicts_with_all(&["addattr", "value"]),
```
jonner, line 156: This is wrong. `--index` can be used with addattr to insert a new attribute at a specific location.

fiuczy: ups, forgot about that combination. Will fix it shortly.

fiuczy: The new version allows the `--index` with `--addattr` combination as well.

src/cli.rs
```diff
@@ -117,7 +117,6 @@ pub enum MdevctlCommands {
                 Running devices are unaffected by this command.",
         group(
             clap::ArgGroup::new("modify")
-                .required(true)
                 .args(&["addattr", "delattr", "jsonfile"]),
```
jonner, line 120: So, this `ArgGroup` is now basically unnecessary, as far as I can tell. The only thing it now does is ensure that only one of `addattr`, `delattr`, or `jsonfile` is specified, but we already have these as explicit requirements down below (using `conflicts_with()`, etc).

So as far as I can tell, this change is essentially equivalent to reverting 86793726dde0156c36e80f2065ba6a222a3f9361, which means that we no longer have any required arguments to the `modify` command. For example,  `mdevctl modify -u $UUID` no longer gives any error message. Can we just leave it required and add `auto` and `manual` to the group? It means that we would no longer be able to use e.g. `--auto` along with `--delattr` or `--addattr`, but I'm not sure that ever really made sense anyway.

src/cli.rs
```diff
@@ -137,20 +136,33 @@ pub enum MdevctlCommands {
...
+            help = "Delete an attribute"
+        )]
+        delattr: bool,
+        #[clap(
+            long,
+            short,
+            requires("index_requires"),
+            help = "Index of the attribute to modify"
+        )]
+        index: Option<u32>,
```
jonner, line 165: The rest of these changes don't have anything to do with the `auto` and `manual` parameters mentioned in the commit message. They should probably be in a separate commit (or commits, since there's a couple independent things happening here).

src/cli.rs
fiuczy: ok, I added `auto` and `manual` to the modify group. That makes the checking of the additional parameters easier.
I pushed an update.

src/cli.rs
fiuczy: With the group the minimum required parameter is specified but the addition parameters are conditional per provided "minimum required parameter". This is what happened and also happens here in the updated version.

PR #92, opened 2023-05-25
tests: use printf over echo

collinwalling:
`echo` + options can run into issues on certain distributions if echo's implementation has been overridden. Though the same case may be present with `printf`, it at least satisfies the Ubuntu-latest CI environment.

Context: the `good-json-null-terminated.sh` file was initially missing its executable privileges, so the GitHub CI gave a false positive. After remedying this, it was apparent that Ubuntu has issues with the `echo -e` flag, though this runs fine on other distros (e.g. fedora). `printf` works for the Ubuntu CI, as well as locally on Fedora.

Fixes 60e377742996 ("trim trailing null from callout script get-attributes output")

### DIFF
--- a/tests/callouts/bad-json.sh
+++ b/tests/callouts/bad-json.sh
@@ -1,2 +1,2 @@
 #!/bin/sh
-echo "not json"
+printf "not json"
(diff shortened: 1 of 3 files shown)


fiuczy (approved):
Tested this patch on fedora and ubuntu successfully.
Reviewed-by: Boris Fiuczynski <fiuczy@linux.ibm.com>

PR #96, opened 2023-07-28
callouts: set script in callout on success only

fiuczy:
The script in callout should not be set with a script when an invocation_failure is returned or the search continues.

### DIFF
--- a/src/callouts.rs
+++ b/src/callouts.rs
@@ -311,9 +311,11 @@ impl<'a, 'b> Callout<'a, 'b> {
                     let r = match self.invoke_first_matching_script(dir, event, action, stdin) {
                         Some((p, o)) => {
                             self.print_err(&o, &p);
-                            self.script = Some(p.clone());
                             match o.status.code() {
-                                Some(0) => Ok(Some(o)),
+                                Some(0) => {
+                                    self.script = Some(p);
+                                    Ok(Some(o))
+                                }
                                 Some(n) => Err(invocation_failure(&p, Some(n))),
                                 None => continue,
                             }


fiuczy:
polite ping...

jonner:
I'm not necessarily opposed to this, but I'm not sure that it's obviously correct. Can you provide any additional justification for why it's necessary?

`invoke_first_matching_script()` will not actually return a `None` value for `o.status.code()` (see line 41 above), so although we handle that case in this function for completeness, it's not actually a realistic possibility.

Also, a specific callout script must return a response code of 2 if the script doesn't handle a particular mdev. But `invoke_first_matching_script()` already handles skipping callouts that return a status code of 2. This means that any response code that we get here is either a success (0) or failure (non-2). And by definition any non-2 response code means that this *is* the correct script for this device. So it doesn't seem obviously incorrect to set `self.script` in this case.

fiuczy:
> I'm not necessarily opposed to this, but I'm not sure that it's obviously correct. Can you provide any additional justification for why it's necessary?
> 
> `invoke_first_matching_script()` will not actually return a `None` value for `o.status.code()` (see line 41 above), so although we handle that case in this function for completeness, it's not actually a realistic possibility.

Currently it is true but if at some point in the future code is changed this should be handled properly.

> 
> Also, a specific callout script must return a response code of 2 if the script doesn't handle a particular mdev. But `invoke_first_matching_script()` already handles skipping callouts that return a status code of 2. This means that any response code that we get here is either a success (0) or failure (non-2). And by definition any non-2 response code means that this _is_ the correct script for this device. So it doesn't seem obviously incorrect to set `self.script` in this case.

It seems very strange to me when using `callout` the first time, which returns with an error caused by searching a script which returns with an error code, the failing script gets internally tied to callout object. It is true that `invoke_first_matching_script()` is shielding the script RC=2 but again (see above) this function internal knowledge should not be relied upon as it creates a hard to observe code dependency.

It has been a long time since came across this when trying to code the reuse of this code for the version support. I think that this caused a problem when trying to make `invoke_first_matching_script()` generic enough to reuse it in the version callout scenarios in which the `callout` object makes two attempts (new and old pattern) to find a script. At that point it seemed simply wrong by just looking at the `callout` function code alone. The first round of a `callout` function is looking for a script with version support and inside the generic `invoke_first_matching_script()` an error is returned if no appropriate version supporting script is found. In that case setting the script would prevent the second round using the old pattern to search a script.

So maybe I simply include this back into the version support series.

jonner:
As I said, I'm not opposed to this patch, I just wanted more justification. I don't really see any downside to this change, so I'll just go ahead and merge it.

PR #97, opened 2023-07-28
callouts: a few cleanups

fiuczy:
Splitting off a few patches collected while implementing versioning as they are mostly unrelated.


fiuczy:
polite ping...

PR #98, opened 2023-08-09
main: fix file open error being suppressed

fiuczy:
This was originally introduced by commit f23dea4 and than spread by commit 863a082.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -65,7 +65,7 @@ fn define_command_helper(
 
     if let Some(jsonfile) = jsonfile {
         let _ = std::fs::File::open(&jsonfile)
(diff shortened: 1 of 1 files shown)


fiuczy:
polite ping...

jonner:
This is weird. We do actually report the error later on in the function when `fs::read_to_string(&jsonfile)` fails. Should we just delete this whole early check?

fiuczy:
> This is weird. We do actually report the error later on in the function when `fs::read_to_string(&jsonfile)` fails. Should we just delete this whole early check?

I like the early checks since they are done with the other checks for a particular scenario and are not spread all over the place.
Btw. in `start_command_helper` the error handling is already correctly done.

PR #101, opened 2023-09-28
Callout version live v4

fiuczy:
The first sixes patches are callout version support related.
The fourth new patch adds code preventing repeating callouts when version lookup failed before. This addresses a comment I made in v3 already.
The last six patches add live modify support exploiting the callout version support.

I addressed all v3 review comments.
Most impact was created by refactoring functions to be generic enough for later code reuse.

### DIFF
--- a/mdevctl.rst
+++ b/mdevctl.rst
@@ -468,6 +468,72 @@ the device type.
             }
         ]
 
+``Get-capabilities``
+
+    A get event is invoked on every new mdevctl execution to find a matching script
+    supporting versioning for the device type.
+    Event type is ``get``. Action is ``capabilities``. State is ``none``.
+    Note that, unlike other call-outs events, **get-capabilities provides a
+    versioning JSON on stdin, and a versioning JSON is returned via stdout**.
+    If a script is found the script is used for every event and action for the
+    device type. Should no script be found the none versioning pattern is used.
+
+    If a valid versioning JSON is returned on stdout and the return code is NOT 2
+    the script is a positive match for the provided device type. A script providing
+    versioning is the primary choice for a device type when mdevctl is executing
+    callouts.
+
+    A script is provided on standard in with a versioning JSON describing the mdevctl
+    supported version, actions and events. Example::
+
+        {
+          "provides": {
+            "version": 1,
+            "actions": [
+              "start",
+              "stop",
+              "define",
+              "undefine",
+              "modify",
+              "attributes",
+              "capabilities"
+            ],
+            "events": [
+              "pre",
+              "post",
+              "notify",
+              "get"
+            ]
+          }
+        }
+
+    A script that wants to support versioning must return a versioning JSON on standard
+    output. The script should list all supported actions in the actions array and all
+    supported events in the events array. It is possible to add additional actions or
+    events in the array but if mdevctl did not have these in the arrays in provides
+    they are ignored. Example::
+
+        {
+          "supports": {
+            "version": 1,
+            "actions": [
+              "start",
+              "stop",
+              "define",
+              "undefine",
+              "modify",
+              "attributes",
+              "capabilities"
+            ],
+            "events": [
+              "pre",
+              "post",
+              "notify",
+              "get"
+            ]
+          }
+        }
+
 AUTO-START CALL-OUTS
 --------------------
 
--- a/mdevctl.rst
+++ b/mdevctl.rst
@@ -41,7 +41,8 @@ The following options are understood:
 
 ``-d|--defined``
     List all defined devices, even if not active. Valid for the ``list``
-    command.
+    command. Modify the defined configuration of a device, even if the
+    device is active. Valid for the ``modify`` command.
 
 ``--delattr``
     Delete an attribute entry. Valid for the ``modify`` command.
@@ -57,6 +58,10 @@ The following options are understood:
 ``--jsonfile=FILE``
     Read the configuration for a device from a JSON file *FILE*.
     Valid for the ``define`` and ``start`` commands.
+
+``-l|--live``
+    Modify active device without modifying the defined configuration of
+    the device. Valid for the ``modify`` command.
 
 ``-m|--manual``
     Do not start a device automatically on parent availability. Valid
@@ -111,8 +116,12 @@ The following commands are understood:
     Attributes can be added or deleted. Attributes to be deleted must be
     specified by their index; if an attribute is specified without an
     index, it is appended at the end of the attribute list.
-    Running devices are unaffected by this command; changes in the configuration
-    are applied the next time the device is started.
+    Active devices are unaffected by this command; changes in the configuration
+    are applied the next time the device is started. Depending on installed
+    callout scripts active devices can be modified. With ``-l|--live``
+    modifications can be applied to active devices if a callout scripts supports
+    the event ``live``. The option ``-d|--defined`` also direct the modification
+    to the started device configuration.
 
 ``start`` *DEVICESPEC*
     Start a mediated device. This command can be used to start either a
@@ -352,12 +361,14 @@ file for define, or activating a device for start).
 Essentially, the procedure in mdevctl looks like this:
 
     - command-line parsing & setup
+    - invoke live-command call-out [1]_
     - invoke pre-command call-out
-    - primary command execution [1]_
-    - invoke post-command call-out [1]_
+    - primary command execution [2]_
+    - invoke post-command call-out [2]_
     - invoke notifier
 
-    .. [1] skipped if step 2 fails.
+    .. [1] executed only if live update is requested.
+    .. [2] skipped if step ``invoke pre-command call-out`` fails.
 
 EVENT SCRIPTS
 -------------
@@ -378,7 +389,7 @@ progress, and the mediated device. The parameters are as follows:
 
 ``-e=``\ *event*
     Event type of call-out that is invoked. For call-out scripts, this may be
-    ``pre``, ``post``, or ``get``. For notification scripts, this will
+    ``pre``, ``live``, ``post``, or ``get``. For notification scripts, this will
     always be ``notify``.
 
 ``-a=``\ *action*
@@ -399,8 +410,8 @@ progress, and the mediated device. The parameters are as follows:
 CALL-OUT EVENT SCRIPTS
 ----------------------
 
-A call-out event script is invoked during a ``pre``, ``post``, or ``get``
-event. mdevctl will attempt each script stored in the mdevctl callouts
+A call-out event script is invoked during a ``live``, ``pre``, ``post`` or
+``get`` event. mdevctl will attempt each script stored in the mdevctl callouts
 directory until either a script that satisfies the device type is found or all
 scripts have been attempted. A device script must check the "TYPE" parameter to
 ensure the specified device type is supported, otherwise error code 2 should be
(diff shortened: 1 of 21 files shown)


src/callouts.rs
```diff
@@ -81,9 +171,257 @@ impl Display for State {
...
+        match callout.callout(
+            Event::Get,
+            Action::Capabilities,
+            Some(&json_ce_ver),
+            &CapabilitiesCheckProcessOutput,
+        ) {
+            Ok(op) => match op {
+                Some(_) => {
+                    self.callouts.push(callout.script.clone().unwrap());
+                    callout.script
```
jonner, line 337: So, if we got here, we know that we found a script that replied with version information, but we haven't made any decisions based on that version information. If, in the future, we find a callout script that replies with a version that we don't know about, we will blindly select that script and use it even if there is a better one available.

src/callouts.rs
```diff
@@ -81,9 +171,257 @@ impl Display for State {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct CalloutScript {
```
jonner, line 175: The naming of all of these objects is getting a bit difficult to keep straight at a glance. I think this is a little too similar to `Callout` so I'd prefer something a bit more expressive like `CalloutScriptInfo` or `CalloutScriptMetadata`

src/callouts.rs
```diff
@@ -81,9 +171,257 @@ impl Display for State {
...
+
+    fn lookup_callout_script(&self, parent: &str, mdev_type: &str) -> Option<CalloutScript> {
+        for cs in self.callouts.iter() {
+            if cs.mdev_type.eq_ignore_ascii_case(mdev_type)
+                && cs.parent.eq_ignore_ascii_case(parent)
+            {
+                return Some(cs.clone());
+            }
+        }
+        None
```
jonner, line 277: Is there a reason you didn't use something like a HashMap to store the list of cached scripts for easier lookup?

src/callouts.rs
```diff
@@ -81,9 +171,257 @@ impl Display for State {
...
+impl AsRef<Path> for CalloutScript {
+    fn as_ref(&self) -> &Path {
+        &self.path
+    }
+}
+
+#[derive(Debug)]
+pub struct CalloutScripts {
+    callouts: Vec<CalloutScript>,
+}
```
jonner, line 243: Similar to my comment above about the `CalloutScript` object, I'd rather this be more expressive so it was understandable at a glance. Maybe something like `CalloutScriptCache`.

src/callouts.rs
```diff
@@ -81,9 +171,257 @@ impl Display for State {
...
+            if cs.mdev_type.eq_ignore_ascii_case(mdev_type)
+                && cs.parent.eq_ignore_ascii_case(parent)
+            {
+                return Some(cs.clone());
+            }
+        }
+        None
+    }
+
+    pub fn find_script(&mut self, dev: &MDev) -> Option<CalloutScript> {
```
jonner, line 280: The name is a bit misleading as it will only find a versioned script, not any script.

src/callouts.rs
```diff
@@ -81,9 +171,257 @@ impl Display for State {
...
+                    warn!("Callout script {:?} provides unknown Action type", p);
+                }
+                if cv.has_event(Event::Unknown) {
+                    warn!("Callout script {:?} provides unknown Event type", p);
+                }
+                c.script = Some(CalloutScript::new(
+                    p,
+                    c.dev.parent().unwrap().to_string(),
+                    c.dev.mdev_type().unwrap().to_string(),
+                    cv,
```
jonner, line 413: we're setting `Callout::script` here, and then we're passing the process output back up to the caller, which is `Callout::callout()`. `Callout::callout()` passes that output up to _its_ caller, which is `CalloutScripts::find_script()`. That function then ignores the output, but extracts `Callout::script` from the `Callout` object and passes it back up to _its_ caller, which is `Callout::find_callout_script()`. That function also passes the script object back up to its caller (e.g. `invoke_modify_live()`), which uses the return value to set the value of `Callout::script`. So we're returning the value of `Callout::script` up about 3 levels in order to set `Callout::script` again. This all seems a bit convoluted and difficult to follow.

jonner (commented):
Sorry for the delay in review. Here's my initial review.

mdevctl.rst
```diff
@@ -468,6 +496,72 @@ the device type.
             }
         ]
 
+``Get-capabilities``
+
+    A get event is invoked on every new mdevctl execution to find a matching script
+    supporting versioning for the device type.
+    Event type is ``get``. Action is ``capabilities``. State is ``none``.
+    Note that, unlike other call-outs events, **get-capabilities provides a
+    versioning JSON on stdin, and a versioning JSON is returned via stdout**.
```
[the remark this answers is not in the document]
fiuczy: The idea is that mdevctl provides its capabilities to the script and the scripts can derive their supported features based on that and send it back to mdevctl. Kind of like a hand shake but without the use of a version number as a version number can be problematic in the distros when backporting happens and scripts being shipped with other packages than mdevctl.

Use of the mdevctl versioning JSON is of course just an offer to the scripts and they do not have to follow that pattern. It is also a valid approach for the script to have a hard coded response. The mdevctl version code is capable of handing missing as well as unknown entities in the JSON response.

I will add more information into the description.

src/callouts.rs
```diff
@@ -59,11 +73,87 @@ impl Display for Action {
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CalloutExchange {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    provides: Option<CalloutVersion>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    supports: Option<CalloutVersion>,
+}
```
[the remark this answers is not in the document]
fiuczy: This is actually the answer to your questions with regard to the JSON requirements.
The structure is used to generate the input and also to parse the output of the scripts.
supports is not included in the scripts input and provides is optional in the scripts response.
If I remember correctly I need a wrapper structure around CalloutVersion for the input as well as the output and code/structure reuse seemed obvious here instead of 
```
pub struct CalloutExport {
    #[serde(skip_serializing_if = "Option::is_none")]
    provides: Option<CalloutVersion>,
}
```
and
```
pub struct CalloutImport {
    #[serde(skip_serializing_if = "Option::is_none")]
    provides: Option<CalloutVersion>,
    #[serde(skip_serializing_if = "Option::is_none")]
    supports: Option<CalloutVersion>,
}
```

src/callouts.rs
fiuczy: Hmm... this is an interesting thought. Can you define "better" in this scenario?
Actually callout script search was never looking beyond the first successful response.
We also took action that we have a reliable order when searching.
I think that this is a separate complex feature that should not be mangled into versioning.
If this would be implemented the decision making of what is better or worse needs to be clearly described to the script providers/writers.

src/callouts.rs
fiuczy: OK, I will rename `CalloutScript` into `CalloutScriptInfo`.

src/callouts.rs
fiuczy: OK, I renamed `CalloutScripts` into `CalloutScriptCache`.

src/callouts.rs
fiuczy: Well, I do not expect that there will be a large number of callout scripts (e.g. 20+) and therefore the performance difference of Vec vs hashmap is not of relevance. Actually according to this https://gist.github.com/daboross/976978d8200caf86e02acb6805961195 it is more in favor of Vec for low numbers.

src/callouts.rs
fiuczy: I agree that is complicated. It is caused by the `Environment` instance being the owner of the `CalloutScriptCache` singleton and this singleton making a reuse of `Callout` by initializing another `Callout` instance.
`invoke_modify_live()` therefore sets on success the script on another `Callout` instance the first time an not again!
So I guess that "ease of reuse" can also create complex scenarios.

src/callouts.rs
```diff
@@ -81,9 +171,257 @@ impl Display for State {
...
+            None => Err(anyhow!(
+                "Output of callout script {:?} is not a valid capabilities XML response",
+                p
+            )),
+        }
+    }
+
+    fn process(&self, c: &mut Callout<'_, '_>, p: PathBuf, o: Output) -> Result<Option<Output>> {
+        c.print_err(&o, &p);
+        match CalloutScripts::parse_script_capabilities(&o) {
```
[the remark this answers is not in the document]
fiuczy: `check` allows to generically implement how INSIDE of `invoke_first_matching_script` the validation the output of a script that matches the device type (or RC!=2) is done.
`process` allows to generically implement how in the search loop in method `callout` behaves.

`DefaultCheckProcessOutput` contains the code for all callouts but `get-capabilities`. It should be most similar to what the code was like before when removing the interfaces by inlining the code.
`CapabilitiesCheckProcessOutput` is used in `get-capabilities` only. The parsing is done twice because as explained above the code gets executed in different levels during execution and I wanted to keep check (or validate) and process (or doing) independent from another so future scenarios are potentially covered as well.

`get-attributes` operates in different mode than `get-capabilities`. `get-capabilities` is a new script event-action which is used to identify if a scripts supports the new "versioning" support. Therefore its behavior does not only rely on the scripts return code but also on the returned output and if that output can be parsed and contains versioning data and based on that a search for a script stops or continues.
`get-attributes` does not do script searching and decides if a script is valid or not.

src/callouts.rs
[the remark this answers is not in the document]
jonner: To me re-using the struct here makes things less clear. And also, the second struct only needs the `supports` field. Your documentation doesn't mention anything about the script returning "provides". So it would instead be something like this, I think.
```
pub struct CalloutExport {
    provides: Option<CalloutVersion>,
}
pub struct CalloutImport {
    supports: Option<CalloutVersion>,
}
```
No need for the skip_serializing stuff anymore either.

mdevctl.rst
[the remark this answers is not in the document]
jonner: I think this stuff has already gotten too complex, to be perfectly honest. We just need a simple protocol negotiation that's well-specified and not left open to interpretation. So the documentation should specify what the script is expected to do with the information and how to respond, in my opinion.

src/callouts.rs
jonner: If there is no discriminating between "better" matches and we just choose the first script that responds anything, then what's the point of all of this complicated versioning info negotiation? We could do almost all of this with a simple version number or yes/no answer, couldn't we?

src/callouts.rs
jonner: I wasn't discussing performance, which is definitely not relevant here. I as discussing code legibility, etc.

For example, compare: 
```rust
   for cs in self.callouts.iter() {
            if cs.mdev_type.eq_ignore_ascii_case(mdev_type)
                && cs.parent.eq_ignore_ascii_case(parent)
            {
                return Some(cs.clone());
            }
        }
        None
    }
```
vs:
```rust
    self.callouts.get(&(parent, mdev_type));
```

src/callouts.rs
jonner: Fair enough, I apparently missed the fact that these are two diferent Callout instances.

src/callouts.rs
fiuczy: Versioning support on its core is independent from a functional support aspect.
It is just a concept to detect that a script supports the new `get-capabilities` interface.
Using in the return just a simple version number as an indicator is worthless as it does not provide the information what a script has implemented. It can response with version 2 and has action attributes not implement which causes mdevctl to make the callout as version 2 suggests the script has it implemented. With that we kind of end up in the same situation as before: try it out and something went wrong. The only way to deal with that is to have again a stdin/stdout parsable contract as implemented for `get-capabilities` for EVERY new (version>1) action/event.
Also there might be even newer scripts than mdevctl which e.g. return hardcoded version 3. Should mdevctl not use that script or assume it backwards compatible since no action/event deprecation will ever happen in newer versions?

Therefore I thought it would be better that the script explains what actions and events it supports and make the callouts just if they are supported by the script. This also enables mdevctl to report to the user if what he tries to do is possible/supported or not. Please check how it is used later for event ``live``.

The selection of which script is selected (besides the versioning) has not changed. It remains as is based on the sorting of the scripts file names and the path.

src/callouts.rs
fiuczy: OK, I will check it out changing the `Vec` into `HashMap`.

fiuczy: OK, the one liner is reformatted by `cargo fmt` into
```
        self.callouts
            .iter()
            .find(|cs| {
                cs.mdev_type.eq_ignore_ascii_case(mdev_type)
                    && cs.parent.eq_ignore_ascii_case(parent)
            })
            .cloned()
```
This is not very different from the for loop which reads to me better but most likely is not rustified enough.

fiuczy:
Closing this PR as PR #108 Callout version live v6 has been merged

PR #107, opened 2023-12-22
Cargo.toml: move tempfile to dev-dependencies

athos-ribeiro:
This is only used as a test dependency. There is no need to pull this in for package builds. Moreover, it should not be listed as a dependency.


jonner:
Thanks!

PR #112, opened 2024-01-31
callouts: protect against panic in callout

fiuczy:
Check mdev type exists before initializing a callout instance to protect against a panic.


fiuczy:
I created https://github.com/mdevctl/mdevctl/pull/113 and therefore close this PR

PR #116, opened 2024-04-12
prevent panic when parsing attributes in JSON

fiuczy:
Using an invalid json causes a panic if an attribute value is not of type string.

Example:
{"mdev_type": "vfio_ap-passthrough", "start": "manual", "attrs":[{"assign_adapter": 0},{"assign_domain": 4}]}

mdevctl define --parent matrix --uuid eeeeeeee-ffff-1111-2222-333333333333 --jsonfile invalid.json thread 'main' panicked at src/mdev.rs:162:51:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


src/mdev.rs
```diff
@@ -159,8 +159,14 @@ impl MDev {
...
-                        let valstr = val.as_str().unwrap();
-                        self.attrs.push((key.to_string(), valstr.to_string()));
+                        if let Some(valstr) = val.as_str() {
+                            self.attrs.push((key.to_string(), valstr.to_string()));
+                        } else {
+                            return Err(anyhow!(
+                                "invalid JSON format for attribute {{{:?}, {}}}: value must be of type str",
+                                key, val
+                            ));
+                        }
```
jonner, line 169: This is fine, but I'd personally prefer something like 
```suggestion
                        let valstr = val.as_str().ok_or_else(|| anyhow!(...))?;
                        self.attrs.push((key.to_string(), valstr.to_string()));
```

fiuczy: Changed.

jonner:
Thanks

PR #117, opened 2024-04-30
60-mdevctl.rules: close FD 3 explicitly

mhartmay:
It's good practice to close file descriptors explicitly.


mhartmay:
Hmm, I'm not sure if my code is correct and that I fully understand what's going on here...

`/bin/sh -c '{ /usr/sbin/mdevctl start-parent-mdevs %k 2>&3 | logger -t mdevctl; } 3>&1 1>&2 3>&- | logger -t mdevctl -p 2'`
So we duplicate the FD3 into FD2, but FD3 is not opened yet? Or who opens FD3?

mhartmay:
Okay my bad, it works as expected and my proposed change doesn't.

```
/bin/sh -c '{ /usr/sbin/mdevctl start-parent-mdevs %k 2>&3 3>&- | logger -t mdevctl; } 3>&1 1>&2 | logger -t mdevctl -p 2'
```
Should work (thanks to Joy).

PR #119, opened 2024-05-06
Fixes stop and refactors sysfs loading

fiuczy:
The original reason to start this series was to reorganize the loading of the data from the sysfs and separate the loading, error handling and data handling. At the end the data is retrieved into a new structure, the places which require data loading from sysfs decide how to handle errors and what is done with the retrieved data.

Besides that a few other things crossed my path and got changed:
 - adds tests for corrupted mdevs
 - adds checks for testdata dead files
 - contains a few rustification patches
 - tries to handle a race when loading sysfs data and a parallel delete of the mdev


fiuczy:
I added another patch which fixes clippy warnings in the tests code and resolved the resulting merge conflicts as well as added changes in the new code to remove clippy warnings.

src/tests/mod.rs
```diff
@@ -247,27 +247,21 @@ impl TestEnvironment {
...
 
 fn regen(filename: &PathBuf, data: &str) -> Result<()> {
     let parentdir = filename.parent().unwrap();
     fs::create_dir_all(parentdir)?;
 
     fs::write(filename, data.as_bytes())
-        .and_then(|_| {
-            println!("Regenerated expected data file {:?}", filename);
-            Ok(())
-        })
```
jonner, line 270: I think maybe we just remove the whole function chaining here instead:

```rust
    fs::write(filename, data.as_bytes())?;
    println!("Regenerated expected data file {:?}", filename);
    Ok(())
```

jonner (commented):
First patch looks fine with one minor suggestion. I don't have these clippy warnings so you must have been using a different version of clippy.

src/tests/mod.rs
fiuczy: Ok, changing it.

src/tests/modify.rs
```diff
@@ -673,7 +667,9 @@ fn test_modify() {
     test_modify_helper(
         "live-fail-without-jsonfile",
         Expect::Fail(Some(
-            format!("'live' option must be used with 'jsonfile' option").as_str(),
+            "'live' option must be used with 'jsonfile' option"
+                .to_string()
+                .as_str(),
```
fiuczy, line 678: I will remove lines 671 and 672.

fiuczy:
> First patch looks fine with one minor suggestion. I don't have these clippy warnings so you must have been using a different version of clippy.

You should be able to get to the messages with `cargo clippy --all-targets --all-features -- -D warnings`.

jonner (commented):
For patch 2 ("add tests with broken active mdevs"), I'm a little bit confused. Did you actually run into broken mdevs in sysfs during normal operation?  And if so, was mdevctl doing the wrong thing in this case? This seems to only add tests cases and doesn't seem to modify behavior...

src/mdev.rs
```diff
@@ -102,14 +102,25 @@ impl MDev {
 
     pub fn load_from_sysfs(&mut self) -> Result<()> {
         debug!("Loading device '{:?}' from sysfs", self.uuid);
+        let parentname = || -> Result<String> {
+            let canonpath = self.path().canonicalize()?;
+            let sysfsparent = canonpath
+                .parent()
+                .ok_or_else(|| anyhow!("Path to parent of mdev {:?} does not exist", self.uuid))?;
+            canonical_basename(sysfsparent)
+        };
```
jonner: These closures seem unnecessary and indirect. We could simply replace the `unwrap()` call below directly with an `ok_or_else()?` and achieve the same thing with a single line change.

jonner (commented):
In the patch "add check in load_from_sysfs for parallel device removal":

We might as well just remove the initial `path.exists()` check if we're just going to check it again anyways. And in fact, I'd prefer to not call `Path::exists()` at all. `Path::canonicalize()` already traverses all symlinks, so I think we can simply match against the error returned by `Path::canonicalize()` and check if it is a `std::io::ErrorKind::NotFound`.

jonner:
I see now that the behavior tested in patch 2 was modified by later commits...

jonner:
To me, this `MDevSysfsData` object feels a bit awkward. It feels like it would be nicer if the `MDev` object had separate 'active config' and a 'defined config' members rather than having this standalone object that can be queried and then used to set values on an `MDev` object. Did you try something like that?

src/mdev.rs
```diff
@@ -16,6 +16,94 @@ pub enum FormatType {
...
 
+pub struct MDevSysfsData {
+    pub uuid: Uuid,
+    pub active: bool,
+    pub parent: Option<String>,
+    pub mdev_type: Option<String>,
+}
+
+impl MDevSysfsData {
+    pub fn new(env: &Rc<dyn Environment>, uuid: &Uuid) -> Result<MDevSysfsData> {
```
jonner: to me, `new()` indicates a simple allocation, not a bunch of I/O that can fail. So something like `load()` would be a much better name.

src/mdev.rs
```diff
@@ -16,6 +16,94 @@ pub enum FormatType {
     Defined,
 }
 
+pub struct MDevSysfsData {
+    pub uuid: Uuid,
+    pub active: bool,
+    pub parent: Option<String>,
+    pub mdev_type: Option<String>,
```
jonner, line 23: I don't think it makes sense to have an `active` field here or `Option`s for the other fields. 

```suggestion
    pub parent: String,
    pub mdev_type: String,
```

If the device is not active, the `load()` function it should simply not return a `MDevSysfsData` object. Having all of these fields be separately nullable implies that it might be valid to have an object where `active` is true, and `parent` is `None`, etc.

src/mdev.rs
```diff
@@ -16,6 +16,94 @@ pub enum FormatType {
...
+    pub fn new(env: &Rc<dyn Environment>, uuid: &Uuid) -> Result<MDevSysfsData> {
+        let mut parent: Option<String> = None;
+        let mut mdev_type: Option<String> = None;
+        let active_path = Self::active_path(env.clone(), uuid);
+        let mut active = Self::is_active(&active_path);
+        if active {
+            parent = match Self::load_parent_from_sysfs(&active_path) {
+                Ok(parentname) => Some(parentname),
+                Err(e) => {
+                    if Self::is_active(&active_path) {
```
jonner: As I said in an eralier comment, checking for `is_active()` before, then calling `load_parent_from_sysfs()`, and then calling `is_active()` again is totally unnecessary. the `load_parent_from_sysfs()` function should already return an error if the file doesn't exist, so you should just be able to call that directly and then check the error type.

src/main.rs
```diff
@@ -413,7 +427,23 @@ fn start_command(
...
+        Err(e) => {
+            if !force {
+                return Err(e);
+            }
+            warn!(
+                "For device {} a sysfs update caused the error: {:?}",
+                dev.uuid, e
+            );
+        }
+    };
```
jonner: This is about the 3rd time I've seen this same basic pattern repeated. It feels a bit awkward and verbose.

src/mdev.rs
```diff
@@ -16,6 +16,94 @@ pub enum FormatType {
...
+        }
+        Ok(MDevSysfsData {
+            uuid: uuid.to_owned(),
+            active,
+            parent,
+            mdev_type,
+        })
+    }
+
+    pub fn new_with_mdev(mdev: &MDev) -> Result<MDevSysfsData> {
```
jonner: same comment about naming: `new` vs. `load`

fiuczy:
> For patch 2 ("add tests with broken active mdevs"), I'm a little bit confused. Did you actually run into broken mdevs in sysfs during normal operation? And if so, was mdevctl doing the wrong thing in this case? This seems to only add tests cases and doesn't seem to modify behavior...

Yes, patch 2 adds only tests which cover the scenario that the mdev is somehow in an incorrect state.
There are no changes in mdevctl behavior itself. These changes are introduced in later patches.

We have seen problems with the mdevctl code which reads data from sysfs when testing with many mdevs.
The scenarios causing errors resulted from mdevctl reading sysfs for mdev x and in parallel the mdev x is deleted.
We have not found mdevctl creating invalid mdevs.

src/mdev.rs
fiuczy: This happens in later patches as this patch is a preparation for further changes.

src/mdev.rs
fiuczy: It seems like it is unnecessary but it is not.
The first `is_active()` is done to check right before getting the parent data from the sysfs.
In case the `load_parent_from_sysfs()` fails this could be for multiple reasons BUT if in parallel the mdev is deleted between the first `is_active()` and the `load_parent_from_sysfs()` this should not cause an error but instead simply be handled as "device is inactive". Same is true for the following `load_mdev_type_from_sysfs()` call.
We have seen this happen in our CI for a few month now.

src/mdev.rs
fiuczy: OK

src/mdev.rs
fiuczy: Ok, I will rename it.

src/mdev.rs
fiuczy: See my comment on `new()`. 
`active` can be `false` and `parent `as well as m`dev_type` are than `None`.
Maybe the struct fields should not be public and only accessible via getters.

src/main.rs
fiuczy: Sure, error handling has a basic pattern but the details differ in every case a bit.

src/mdev.rs
```diff
@@ -118,8 +118,28 @@ impl MDev {
             debug!("device did not exist in sysfs: {:?}", self);
```
fiuczy: > In the patch "add check in load_from_sysfs for parallel device removal":
>
> We might as well just remove the initial path.exists() check if we're just going to check it again anyways. And in fact, I'd prefer to not call Path::exists() at all. Path::canonicalize() already traverses all symlinks, so I think we can simply match against the error returned by Path::canonicalize() and check if it is a std::io::ErrorKind::NotFound.

I can give it a try as I am not sure if the error does match only the link to the devices directory no longer exists. I am also not sure if this improves readability but let me try it out.

fiuczy: I added patch [mdev: MDevSysfsData.load() without path.exists() check](https://github.com/mdevctl/mdevctl/pull/119/commits/96812712ac6fcc74d944614fe6714af17efa6c63) to address this.

src/mdev.rs
fiuczy: In patch [mdev: MDevSysfsData.load() without path.exists() check](https://github.com/mdevctl/mdevctl/pull/119/commits/96812712ac6fcc74d944614fe6714af17efa6c63) `is_active()` gets removed.

fiuczy:
> To me, this `MDevSysfsData` object feels a bit awkward. It feels like it would be nicer if the `MDev` object had separate 'active config' and a 'defined config' members rather than having this standalone object that can be queried and then used to set values on an `MDev` object. Did you try something like that?

I do not think that the raciness problems retrieving consistent data from the sysfs is directly related to how the `MDev` object manages the persisted and active configurations. I do not see why that would be necessary but I suggest to address that issue separately.

fiuczy:
It has been a while since I last push changes. How do we want to continue with this?

fiuczy:
> > For patch 2 ("add tests with broken active mdevs"), I'm a little bit confused. Did you actually run into broken mdevs in sysfs during normal operation? And if so, was mdevctl doing the wrong thing in this case? This seems to only add tests cases and doesn't seem to modify behavior...
> 
> Yes, patch 2 adds only tests which cover the scenario that the mdev is somehow in an incorrect state. There are no changes in mdevctl behavior itself. These changes are introduced in later patches.
> 
> We have seen problems with the mdevctl code which reads data from sysfs when testing with many mdevs. The scenarios causing errors resulted from mdevctl reading sysfs for mdev x and in parallel the mdev x is deleted. We have not found mdevctl creating invalid mdevs.

To give you an idea what we encounter sometimes in our tests:
2024-04-10 13:46:00.959+0000: 8380: error : virMdevctlModify:958 : internal error: Unable to modify mediated device: ap-check: vfio-ap module does not support ap_config for live modificationError: Script '"/usr/lib/mdevctl/scripts.d/callouts/ap-check"' failed with status '255'
2024-04-10 13:46:02.080+0000: 8443: error : nodeDeviceUpdateMediatedDevices:1923 : internal error: failed to query mdevs from mdevctl: Error: Device must have a defined mdev_type
2024-04-10 13:46:02.080+0000: 8443: warning : udevRemoveOneDeviceSysPath:1631 : Update of mediated device failed
2024-04-10 13:46:02.130+0000: 8399: error : nodeDeviceUpdateMediatedDevices:1923 : internal error: failed to query mdevs from mdevctl: Error: Device must have a defined mdev_type

As you can see the problem is not obvious but by fixing the race condition when reading from sysfs this problem is fixed.

fiuczy:
Ping??

jonner:
sorry for the long delay. Merged now.

PR #122, opened 2024-09-10
Improve failure message for live update

jonner:
Resolves: https://issues.redhat.com/browse/RHEL-40353


src/callouts.rs
```diff
@@ -449,7 +449,7 @@ impl<'a> Callout<'a> {
             // live is only supported when script with versioning exists
             debug!("No callout script with version support found that supports live modify");
             return Err(anyhow!(
-                "No callout script with version support found that supports live modify"
+                "Live modification is not supported for this device type. Changing configuration requires stopping and restarting the device."
```
fiuczy: This removes the hint to the user of mdevctl that a callout script needs to exist which also support the new versioning concept in order to support live modify. It needs to be pointed out that the device driver of the device also needs to support a live update/modify concept which the script can make use of. Maybe a hint to some mdevctl documentation could be of help here to keep the message short?

When reading the message the first time I misunderstood the second sentence by "configuration" being the mdevctl configuration. Maybe changing it to "Changing the device configuration requires stopping and restarting the device."

jonner: As demonstrated in the referenced JIRA bug, this error message may be passed up through libvirt and presented to a user who has no idea that anything named `mdevctl` even exists. And this user most likely cannot even do anything about the error. A system administrator will need to install callout scripts for the device if they exist. 

So I think that leaving the detailed information in the `debug!` call will be useful to people who know what they're doing and are trying to debug the issue (maybe we should upgrade it to a `warn!`?). But I think we should make the error message less technical and less confusing to end users by not referencing implementation details.

I think your wording suggestion is good.

fiuczy: I agree with you. 
Wrt `warn!` instead of `debug!`: I think that the error is enough indication to have a look into the debug logs if life modification is supposed to work for a device.

fiuczy:
```suggestion
                "Live modification is not supported for this device type. Changing the device configuration requires stopping and restarting the device."
```

fiuczy (changes requested):
You can add my RB with the suggestion applied.

fiuczy (approved):
Not sure why all for checks failed but the changes look good to me.
Reviewed-by: Boris Fiuczynski <fiuczy@linux.ibm.com>

jonner:
Must have been a temporary server failure. I just re-ran them and they all succeeded.

fiuczy:
Just saw that and was wondering how that happened.

